PR #2, opened 2024-10-06
add the projectKey and the organisationKey for SonarCloud

MaxCastelCast:
Change the app build.gradle so that the right projectKey and organisation key. This enables us to use SonarCloud.
Modify some settings in the SonarCloud project like switching off Automatic Analysis so that the continues integration works normaly


louiscrc (approved):
LGTM

PR #4, opened 2024-10-07
chore: update package name and add Firebase web client ID

JanStaszewiczEPFL:
- Changed package name to `com.github.se.icebreakrr`
- Added Firebase web client ID for authentication setup


MaxCastelCast (approved):
This were necessary changes from the android sample we got, i approve

PR #5, opened 2024-10-08
feat(auth): add firebase signing configs

JanStaszewiczEPFL:
- Set up Firebase Authentication in gradle files


MaxCastelCast (approved):
Yes I think the modifications in the gradle file are good

PR #17, opened 2024-10-10
Feature/navigation

arthur-mrgt:
#### Summary
This draft pull request introduces several key components related to navigation and UI screens for the Icebreakrr app. The code added focuses on implementing the basic structure for navigation between top-level destinations and setting up foundational UI elements for the app.

#### Key Changes
1. **Bottom Navigation Menu**:
   - Implemented a `BottomNavigationMenu` composable for navigation between top-level destinations like *Around You*, *Profile*, and *Notifications*.
   - Integrated navigation actions to handle tab selection and UI transitions.
   - **Note**: The UI of the `BottomNavigationMenu` is **not final** and may be adjusted to the Figma in the future.

2. **Navigation Actions**:
   - Added `NavigationActions` class to manage navigation between top-level destinations and specific screens.
   - Implemented functionality to navigate back and retrieve the current route.

3. **Top-Level Destinations and Routes**:
   - Created the `TopLevelDestination` data class representing each destination.
   - **Unlike in the bootcamp**, I decided to create a **separate file for `TopLevelDestination`** to better organize and separate logic.
   - Defined routes and screen identifiers for easy reference within the app.

4. **Screens for Key Features**:
   - Added composable screens for *Around You*, *Profile*, and *Notifications* with bottom navigation included.
   - **Note**: The content of these screens is currently just placeholder text and will be updated later.

5. **Main App Structure**:
   - Created the `IcebreakrrApp` composable, defining the navigation graph with routes and screens.
   - Set up navigation between the different sections of the app using `NavHost`.

6. **String Resources**:
   - Added string resources for the new screens and navigation, including translations for *Around You*, *Profile*, and *Notifications*.
   - Added Google client ID for authentication setup.

7. **Dependencies**:
   - Added the `androidx.navigation.compose` dependency for Compose-based navigation.
   - Updated `libs.versions.toml` with the version and reference for `androidx-navigation-compose`.

#### Future updates
- The Auth screen integration is currently a TODO and will be added in future PR.
- For now there is only a simple text placeholder for each screen.


app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -4,40 +4,65 @@ import android.os.Bundle
...
+fun IcebreakrrApp() {
+  val navController = rememberNavController()
+  val navigationActions = NavigationActions(navController)
 
-@Preview(showBackground = true)
-@Composable
-fun GreetingPreview() {
-  SampleAppTheme { Greeting("Android") }
+  // TODO Implement Auth Screen navigation
+  NavHost(navController = navController, startDestination = Route.AROUND_YOU) {
```
SamSteu, line 46: The startDestination should be Route.Auth

app/src/main/java/com/github/se/icebreakrr/ui/navigation/NavigationAction.kt
```diff
@@ -0,0 +1,51 @@
...
+    navController.popBackStack()
+  }
+
+  /**
+   * Get the current route of the navigation controller.
+   *
+   * @return The current route
+   */
+  open fun currentRoute(): String {
+    return navController.currentDestination?.route ?: "unknown"
```
SamSteu, line 49: If the current route is null (which should not happen), we can redirect the user to the auth screen. If the current route is "unknown," the application will likely crash.

SamSteu (commented):
In the Screens for the Settings, Around you and Notifications, it can be useful to use the contentPadding of the scaffold to avoid overlapping between the content and the bottom/top navigation bar .

For the TopLevelDestination, you can name "Profile" into "Settings" instead because in the bottomNavigationMenu you can click on the settings and then, from the settings you can access to your profile. (To be changed also in the MainActivity.kt)

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
MaxCastelCast: I agree with Samuel

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -4,40 +4,65 @@ import android.os.Bundle
...
+      composable(Screen.PROFILE) { ProfileScreen(navigationActions) }
+    }
+
+    navigation(
+        startDestination = Screen.NOTIFICATIONS,
+        route = Route.NOTIFICATIONS,
+    ) {
+      composable(Screen.NOTIFICATIONS) { NotificationScreen(navigationActions) }
+    }
+  }
```
MaxCastelCast, line 67: It isn't clear what "ProfileScreen" means, is it the screen of our own profile or of one we are seeind in the around you screen ?
Also, navigations are supposed to encapsulate feature-specific flows, so we could have the "around you screen" and the "other's profile screen" in the same navigation ?

app/src/main/java/com/github/se/icebreakrr/ui/navigation/NavigationAction.kt
arthur-mrgt: Yes, I initially planned to handle the error elsewhere, but it's better to address it here

MaxCastelCast (approved):
LGTM

PR #20, opened 2024-10-13
Feature/tags UI

MaxCastelCast:
The different funcitons in the TagView file : 
-The dataclass TagStyle which takes a text color, background color and font size 
-The component Tag which takes as argument a text and a TagStyle
-The component ClickTag, a tag on which we can click on
-The component TagSelector, a OutlineTextField which toggles a DropDownMenu with different tags to select, it takes many arguments (documentation in the code)

Some modifications have been done in the build.gradle and the libs.versions files to add more libraries like Accompanist which has a useful UI component, the FlowRow, which is useful for the TagSelector

Here we give different screens where the TagSelector is used : 

Image 1
Image 2
Image 3
Image 4

Now I have added some new features for the tags and how they are displayed :
-Two new composables : RowOfTags and RowOfClickTags that enables us to display a list of tags
-If the list of tags has more than three elements, a ExtendTag is going to show (with the text "..."), and if the user clicks on it, all the tags are displayed
-Added two arguments for the Tag selector : height and width, to decide the size of the TagSelector


louiscrc (changes requested):
I think this is really good work. Though, you should wrap the component in a box with a defined size that we can scroll in, so that this outer component can adapt to the screen size (height especially) and allows user to scroll among the numerous tags.

BotondAKovacs (commented):
Important: you should improve the coverage by adding tests as 66% seems a bit low
Question: isn't it in this file that the "expandable tag list" that we see on the Figma mockup in the profile page should be implemented?
Nitpick: you could put the tag selector in a box and add a parameter to limit the size of the tagselector box

MaxCastelCast:
You are both right, I ll try to add for the TagSelector two arguments, to be able to choose a height and a length. I ll also add more test to have more coverage

louiscrc (approved):

louiscrc:
Nice work Maximo ! Fits perfectly what I was saying in my last comment !

PR #21, opened 2024-10-15
feature/TagsBackend repository

SamSteu:
**Firestore**
The way we store the tags in the firestore is by creating a collection named "Tags". Inside that we have one document per category of tags. In this document we have the name of the category, the color of the tags and a list of subtags (e.g., the document "Sport" contains a color, a name ("Sport") and a list of subtags ("Sport", "Basketball", "Football", ...). The name of the category is present in the subtags because we can use it as a tag),

**Tags**
File that contains : 

- `TagsCategory` which is a data class representing a category of tags (e.g, Sport, Art,...)
  Put in the firestore documents.
- `CategoryString` in an enumeration that contains the different category name

**TagsRepository**
This file make the link between firestore and the code. It contains several functions : 

- `getAllTags` which fetches all the tags from the repository and put them in a `List<TagsCategory>`
- `addTag` that adds a tag to a certain category
- `addCategory` that adds a category 
- `deleteTag` that deletes a tag in a certain category
- `deleteCategory` that deletes an entire category
- `firestoreToTags` which is a utility private functions that transforms a `DocumentSnapshot` into a `TagsCategory`

`addTag`, `addCategory`, `deleteTag`, and `deleteCategory` are functions that won't be used in the code, but it can be useful if we want to scale the tags.

**TagsViewModel**
I just created this file for the moment, it will be populated in my next PR

**MainActivity**
I added the intantiation of the `TagsRepository`

**TagsRepositoryTest**
this file contains tests for every function in `TagsRepository`. At least one for success and one for failure

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
@@ -1,6 +1,22 @@
 package com.github.se.icebreakrr.model.tags
 
-import androidx.compose.ui.graphics.Color
+data class TagsCategory(
+    val uid: Long,
+    val name: String,
+    val color: String,
+    val subtags: List<String>
+)
 
-data class TagsCategory(val uid: Long, val name: String, val color: String, val subtags: List<String>)
-
+enum class CategoryString {
+  Sport,
+  Music,
+  Technology,
+  Travel,
+  Food,
+  Art,
+  Gaming,
+  Cinema,
+  Activism,
+  Occupation,
+  Philosophy
+}
--- a/app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
@@ -1,0 +1,6 @@
+package com.github.se.icebreakrr.model.tags
+
+import androidx.compose.ui.graphics.Color
+
+data class TagsCategory(val uid: Long, val name: String, val color: String, val subtags: List<String>)
+
--- a/app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
@@ -1,10 +1,9 @@
 package com.github.se.icebreakrr.model.tags
 
 data class TagsCategory(
-    val uid: Long,
-    val name: String,
-    val color: String,
-    val subtags: List<String>
+    val name: String = "",
+    val color: String = "#FFFFFFFF",
+    val subtags: List<String> = emptyList()
 )
 
 enum class CategoryString {
--- a/app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
@@ -1,0 +1,41 @@
+package com.github.se.icebreakrr.model.tags;
+
+import android.util.Log
+import com.google.firebase.firestore.DocumentSnapshot
+import com.google.firebase.firestore.FirebaseFirestore
+import kotlin.Unit;
+
+class TagsRepository(private val db: FirebaseFirestore) {
+    private val collectionPath = "Tags"
+    fun getAllTags(onSuccess: (List<TagsCategory>) -> Unit, onFailure: (Exception) -> Unit){
+        val categories: MutableList<TagsCategory> = mutableListOf()
+        db.collection(collectionPath).get()
+            .addOnFailureListener{
+                Log.e("TagsRepository", "[getAllTags] Could not get the tags : $it")
+                onFailure(it)
+            }.addOnSuccessListener { docs ->
+                for (doc in docs.documents){
+                    if (doc.exists()){
+                        firestoreToTags(doc, {categories.add(it)}, {
+                            Log.e("TagsRepository", "[getAllTags] Could not fetch firebase to get the tags : $it")
+                            onFailure(it)
+                        })
+                    }
+                }
+                onSuccess(categories)
+            }
+    }
+
+    private fun firestoreToTags(doc: DocumentSnapshot, onSuccess: (TagsCategory) -> Unit, onFailure: (Exception) -> Unit){
+        try{
+            val uid = doc.getLong("uid") ?: 0
+            val name = doc.getString("name") ?: ""
+            val color = doc.getString("color") ?: "#00000000"
+            val subtags = doc.get("subtags") as? List<String> ?: emptyList()
+            onSuccess(TagsCategory(uid, name, color, subtags))
+        }catch(e: Exception){
+            Log.e("TagsRepository", "[firestoreToTags] Failure in the deserialization of json objects : $e")
+            onFailure(e)
+        }
+    }
+}
--- a/app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
@@ -31,7 +31,7 @@ class TagsRepository(private val db: FirebaseFirestore) {
         .addOnSuccessListener { docs ->
           for (doc in docs.documents) {
             if (doc.exists()) {
-                Log.e("TESTTAGADD", "okay good")
+              Log.e("TESTTAGADD", "okay good")
               firestoreToTags(
                   doc,
                   { categories.add(it) },
@@ -43,7 +43,7 @@ class TagsRepository(private val db: FirebaseFirestore) {
                   })
             }
           }
-            Log.e("TESTTAGADD", "cagtegories : ${categories.size}")
+          Log.e("TESTTAGADD", "cagtegories : ${categories.size}")
           onSuccess(categories)
         }
   }
@@ -106,7 +106,8 @@ class TagsRepository(private val db: FirebaseFirestore) {
    * of subtags. If the category on firebase already exists, add elements of the subcateries that
    * are not in the category already present in the firebase to the category on firebase. Else it
    * just create a new category with the subtags in it. IMPORTANT : the name of the new category
-   * must be put manually after the execution of the function into the enum CategoryString in model/tags/Tags.kt
+   * must be put manually after the execution of the function into the enum CategoryString in
+   * model/tags/Tags.kt
    *
    * @param onFailure : callback function called if there is an error
    * @param categoryName : name of the new category
--- a/app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
@@ -1,19 +1,16 @@
 package com.github.se.icebreakrr.model.tags
 
 import android.util.Log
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
```diff
@@ -0,0 +1,250 @@
...
+    val docRef = db.collection(collectionPath).document(category.name)
+    docRef
+        .get()
+        .addOnFailureListener { exception: Exception ->
+          Log.e("TagsRepository", "[addTag] Could not retrieve the document: $exception")
+          onFailure(exception)
+        }
+        .addOnSuccessListener { documentSnapshot: DocumentSnapshot ->
+          if (documentSnapshot.exists()) {
+            var tagsCategory = TagsCategory("", "#FFFFFFFF", listOf())
```
MaxCastelCast, line 63: Maybe having a by default constructor can be useful to make the code lighter and more readable, since those default values aren't used and are overwritten afterwards

MaxCastelCast (approved):
Looks good to me

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
```diff
@@ -0,0 +1,250 @@
...
+class TagsRepository(private val db: FirebaseFirestore) {
+  private val collectionPath = "Tags"
+
+  /**
+   * Function that get all the tags present in the firestore database
+   *
+   * @param onSuccess : when the operation is successful, returns the list of TagsCategory got from
+   *   the database
+   * @param onFailure : When the operation fails, log an error and returns the onFailure with the
+   *   erorr
```
Kapestian, line 17: typo: `erorr`

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
```diff
@@ -0,0 +1,250 @@
...
+                      "TagsRepository",
+                      "[addCategory] error while converting firebase to Tags : $e")
+                  onFailure(e)
+                })
+            val subtags = tagsCategory.subtags.toMutableList()
+            subtags.addAll(subcategories.filter { it !in subtags })
+            db.collection(collectionPath)
+                .document(categoryName)
+                .update("subtags", subtags.toList())
+                .addOnSuccessListener {}
```
Kapestian, line 137: Question is it normal that most `addOnSuccessListener {}` are empty?

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
```diff
@@ -0,0 +1,250 @@
...
+    if (tagCategory != null && tagCategory.subtags.contains(name)) {
+      db.collection(collectionPath)
+          .document(category.name)
+          .update("subtags", tagCategory.subtags.filter { it != name })
+          .addOnSuccessListener {}
+          .addOnFailureListener { e: Exception ->
+            Log.e("TagsRepository", "[deleteTag] Failed to update the document : $e")
+            onFailure(e)
+          }
+    }
```
Kapestian, line 187: Problem (ft. copilot): there is a problem of mixing synchronous and asynchronous code here. `getAllTags()` is asynchronous as it does network operation. this means `tagList `may not be populated yet if you are using it just after the `getAllTags()` block.

Here is a solution proposed by copilot (to show you how asynchronous code should be used)
```

fun deleteTag(onFailure: (Exception) -> Unit, name: String, category: CategoryString) {
  getAllTags(
      { tagList ->
        val tagCategory = tagList.find { it.name == category.name }
        if (tagCategory != null && tagCategory.subtags.contains(name)) {
          db.collection(collectionPath)
              .document(category.name)
              .update("subtags", tagCategory.subtags.filter { it != name })
              .addOnSuccessListener {}
              .addOnFailureListener { e: Exception ->
                Log.e("TagsRepository", "[deleteTag] Failed to update the document : $e")
                onFailure(e)
              }
        } else {
          Log.d(
              "TagsRepository",
              "[deleteTag] the tag wasn't in the database or the category doesn't exist : $name")
        }
      },
      { e ->
        Log.e("TagsRepository", "[deleteTag] error while getting tags : $e")
        onFailure(e)
      }
  )
}
```

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
```diff
@@ -0,0 +1,250 @@
...
+    var tagList = emptyList<TagsCategory>()
+    getAllTags(
+        { tagList = it },
+        { Log.e("TagsRepository", "[deleteCategory] error while getting tags : $it") })
+    val tagCategory =
+        try {
+          tagList.filter { it.name == category.name }[0]
+        } catch (e: Exception) {
+          null
+        }
```
Kapestian, line 212: Problem: same as `deleteTag()` problem (asynchronicity)

app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
```diff
@@ -0,0 +1,21 @@
+package com.github.se.icebreakrr.model.tags
+
+data class TagsCategory(
+    val name: String = "",
+    val color: String = "#FFFFFFFF",
+    val subtags: List<String> = emptyList()
```
Kapestian, line 6: Question: shouldn't at least `name` be mandatory?

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsViewModel.kt
```diff
@@ -0,0 +1,3 @@
+package com.github.se.icebreakrr.model.tags
+
+data class TagsViewModel(val repository: TagsRepository) {}
```
Kapestian, line 3: Question: why is it empty and unused in the main screen?

app/src/test/java/com/github/se/model/tags/TagsRepositoryTest.kt
```diff
@@ -0,0 +1,415 @@
...
+
+    `when`(task.addOnFailureListener(any())).thenAnswer { invocation -> task }
+
+    repository.addCategory({}, "Chaton", listOf("Kayak", "Trail"), "#FFFFFFFF")
+
+    assertEquals(true, success)
+  }
+
+  @Test
+  fun deleteTagSuccessfullTest() {
```
Kapestian, line 226: typo: `deleteTagSuccessfullTest` -> `deleteTagSuccessfulTest`

app/src/test/java/com/github/se/model/tags/TagsRepositoryTest.kt
```diff
@@ -0,0 +1,415 @@
...
+    `when`(doc2.getString("name")).thenReturn("Music")
+    `when`(doc2.getString("color")).thenReturn("0xFF0000FF")
+    `when`(doc2.get("subtags")).thenReturn(listOf("Rock", "Tech", "Classical"))
+
+    `when`(task.addOnSuccessListener(any())).thenAnswer { invocation: InvocationOnMock ->
+      val listener = invocation.arguments[0] as OnSuccessListener<QuerySnapshot>
+      listener.onSuccess(querySnapshot)
+      task
+    }
+    `when`(task.addOnFailureListener(any())).thenAnswer { invocation: InvocationOnMock -> task }
```
Kapestian, line 75: Problem (ft. copilot): it seems there is a chaining problem in that mocking. As I'm getting really tired I'll let you do the research but here is the copilot message (maybe it is completely bogus): 

> Incorrect Mocking of Task
The issue here is with the mocking of the addOnSuccessListener and addOnFailureListener methods for the Task object. These methods are mocked but not properly chained, which might lead to unexpected behavior. Specifically, the addOnSuccessListener method should return the Task object to allow method chaining, but this is not guaranteed in the provided code.

Kapestian (changes requested):
You did a lot of work, that's impressive. There are some blocking problems (that will be fixed pretty easily), I'll let you make the changes necessary and ask another review when you're done. For the test part, I'm no expert (yet) but I see there are some strange things, you should maybe double-check with an AI assistant to be sure.

As a famous philosopher said: **"Can we fix it? Yes we can!"**

JanStaszewiczEPFL:
I generally agree with @Kapestian comments there is an issue with how the MVVM model is structured and handled in this PR and there is no reason to apply the model only from the next one. Otherwise LGTM.

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
SamSteu: Here you are in the function to add a category. When the operation is successful there is nothing really to return so I decided to don't have a `onSuccess` parameter for this function. So in the onSuccess there is nothing done. If you should have an `onSuccess` parameter and call it at `.addOnSuccessListener` I can change it .

app/src/main/java/com/github/se/icebreakrr/model/tags/Tags.kt
SamSteu: This is just a constructor used in some cases where I want to instantiate a `TagsCategory` so that I can assign to it a value later, so I don't use the name at this moment

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsViewModel.kt
SamSteu: As I said in my PR, it will be in my next PR, if I do the viewModel in this one there will be too much change for a PR.

app/src/test/java/com/github/se/model/tags/TagsRepositoryTest.kt
SamSteu: I thing copilot is completely bogus (I indeed return the task).

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsViewModel.kt
Kapestian: Oh you're right. As repository is not used really in this PR, you could just remove it from mainscreen and replace it with a comment mentioning your intention of adding the VM later

app/src/main/java/com/github/se/icebreakrr/model/tags/TagsRepository.kt
```diff
@@ -0,0 +1,241 @@
...
+                  Log.e("TagsRepository", "[deleteTag] Failed to update the document : $e")
+                  onFailure(e)
+                }
+          }
+          Log.d(
+              "TagsRepository",
+              "[deleteTag] the tag wasn't in the database or the category doesn't exist : $name")
+        },
+        { Log.e("TagsRepository", "[deleteTag] error while getting tags : $it") })
+  }
```
Kapestian, line 187: Nitpick: try expliciting the parameters of getAllTags to make code more readable

```
getAllTags(
   onSuccess=..., 
   onFailure=...
) 
```

Kapestian (approved):
I added one or two comments, but you can ignore them if you want to start your next task

JanStaszewiczEPFL (approved):
LGTM

MaxCastelCast (approved):
LGTM

PR #23, opened 2024-10-16
Feature/filter page UI

BotondAKovacs:
Summary:
Creation of the filter UI page that contains a gender filter, an age range filter and a tag filter.

Key Features:

- Gender selector: the user can select any number of them at the same time 
- Age range specifier: the user can specify an age range, the lower limit is 13, if either isn't an integer or the "to" is smaller than the "from" both text fields get cleared and the user gets a toast
- Tag selector: using the TagView's tag selector, the user can choose/remove any number of tags

Future Updates:

- The backend isn't implemented yet, this is just the UI

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
@@ -15,6 +15,7 @@ import com.github.se.icebreakrr.ui.navigation.NavigationActions
 import com.github.se.icebreakrr.ui.navigation.Route
 import com.github.se.icebreakrr.ui.navigation.Screen
 import com.github.se.icebreakrr.ui.sections.AroundYouScreen
+import com.github.se.icebreakrr.ui.sections.FilterScreen
 import com.github.se.icebreakrr.ui.sections.NotificationScreen
 import com.github.se.icebreakrr.ui.sections.SettingsScreen
 import com.github.se.icebreakrr.ui.theme.SampleAppTheme
@@ -49,6 +50,7 @@ fun IcebreakrrApp() {
         route = Route.AROUND_YOU,
     ) {
       composable(Screen.AROUND_YOU) { AroundYouScreen(navigationActions) }
+      composable(Screen.FILTER) { FilterScreen(navigationActions) }
     }
 
     navigation(
--- a/app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
@@ -101,7 +101,7 @@ fun FilterScreen(navigationActions: NavigationActions) {
   var ageRangeError by remember { mutableStateOf(false) }
 
   fun validateAndUpdateAgeFrom(input: String) {
-      ageFromInput = input
+    ageFromInput = input
     if (input.isEmpty()) {
       ageFrom = null
       ageRangeError = false
@@ -118,7 +118,7 @@ fun FilterScreen(navigationActions: NavigationActions) {
   }
 
   fun validateAndUpdateAgeTo(input: String) {
-      ageToInput = input
+    ageToInput = input
     if (input.isEmpty()) {
       ageTo = null
       ageRangeError = false
--- a/app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
@@ -1,0 +1,324 @@
+package com.github.se.icebreakrr.ui.sections
+
+import android.widget.Toast
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.IntrinsicSize
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.filled.ArrowBack
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButton
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.Text
+import androidx.compose.material3.TopAppBar
+import androidx.compose.material3.TopAppBarDefaults
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.onFocusChanged
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.semantics.selected
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.github.se.icebreakrr.ui.navigation.NavigationActions
+import com.github.se.icebreakrr.ui.navigation.Screen
+import com.github.se.icebreakrr.ui.tags.TagSelector
+
+const val textSizeFactor = 0.3f
+val IcebreakrrBlue: Color = Color(0xFF1FAEF0)
+
+// This file was written with the help of Cursor, Claude, ChatGPT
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,324 @@
...
+                  Text("Filter", color = Color.White, fontSize = buttonTextSize)
+                }
+          }
+        }
+      }
+}
+
+@Composable
+fun GenderButton(selected: Boolean, onClick: () -> Unit, label: String) {
+  val screenWidth = LocalConfiguration.current.screenWidthDp.dp
```
MaxCastelCast, line 298: Wouldn't be better to have the width and height of the button passed as arguments instead of having them hardcoded ? I can be useful if those buttons are used somewhere, like when we ll do a Sign up screen

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,324 @@
...
+        Toast.makeText(context, "Invalid age range", Toast.LENGTH_SHORT).show()
+        ageToText = ""
+        ageFromText = ""
+        return // exit early so that there aren't 3 overlapping toasts as the next 2 would also be
+        // called
+      }
+      ageFromText.isNotEmpty() && ageFrom == null -> {
+        Toast.makeText(context, "Please enter a valid 'From' age (13 or older)", Toast.LENGTH_SHORT)
+            .show()
+        ageFromText = "" // Clear invalid input
```
MaxCastelCast, line 151: here aren't you going to have two overlapping Toasts if you have a null ageFrom and null ageTo ? You could have return argument here like when you are in the ageRangeError case

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,324 @@
...
+        ageToText = "" // Clear invalid input
+      }
+    }
+  }
+
+  Scaffold(
+      topBar = {
+        TopAppBar(
+            title = {
+              Text("Filter", color = Color.White, modifier = Modifier.testTag("FilterTopBarTitle"))
```
MaxCastelCast, line 165: Would be more readable to have a composable funciton outside of the Scaffold that does the TopAppBar ? It is ok to keep it like this right now but Since we use the same TopAppBar for the Filter, the Notifications, the AroundYou and the Settings, we put a standardized TopAppBar later on

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,324 @@
...
+                      .firstOrNull { it.first == tag }
+                      ?.let { newTag -> selectedTags.value += newTag }
+                }
+              },
+              height = tagSelectorHeight,
+              width = tagSelectorWidth,
+              textSize = (tagSelectorHeight.value * tagSelectorTextSizeFactor).sp)
+          Box(modifier = Modifier.fillMaxWidth().padding(16.dp)) {
+            Button(
+                onClick = { navigationActions.navigateTo(Screen.AROUND_YOU) },
```
MaxCastelCast, line 280: You can use the funciton navigationActions.goBack() here, it is more optimal i think

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
MaxCastelCast: This doesn't need to be fixed, it is just a remark

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,324 @@
...
+      } else {
+        ageFrom = null
+        ageRangeError = false
+      }
+    }
+  }
+
+  fun validateAndUpdateAgeTo(input: String) {
+    ageToText = input
+    if (input.isEmpty()) {
```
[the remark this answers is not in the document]
MaxCastelCast: Maybe it isn't an issue, maybe wait for an other review before takling this issue

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -49,6 +50,7 @@ fun IcebreakrrApp() {
         route = Route.AROUND_YOU,
     ) {
       composable(Screen.AROUND_YOU) { AroundYouScreen(navigationActions) }
+      composable(Screen.FILTER) { FilterScreen(navigationActions) }
```
MaxCastelCast, line 53: Remark : There is also a button filter in the Notification page, so idk if filter is considered in the workflow of the AroundYou Screen or the notification screen, we can keep it like this for now

MaxCastelCast (commented):
The code is globaly good, the coverage of 88% is very nice, i have however a fiew requests about the lines down below, if you disagree with me about remarks, we can discuss. But all fixes should be quick to do

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
BotondAKovacs: We'll have to talk about it with the whole team, I'm not sure the filter will be available on the inbox page.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
BotondAKovacs: You're right. Thanks.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
BotondAKovacs: You're right, it's temporary.

MaxCastelCast (approved):
LGTM

louiscrc (approved):
Your implementation is very good.
Please notice that you have sometimes use fixed padding, and sometimes made it proportional to the screen size. I think it would be better to generalize the proportional version, for complete reactivity on different screen sizes.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,323 @@
...
+            Pair("psto", Color.Green),
+            Pair("pest", Color.Green),
+            Pair("peest", Color.Green)))
+  }
+
+  val stringQuery = remember { mutableStateOf("") }
+  val expanded = remember { mutableStateOf(false) }
+
+  var ageFromText by remember { mutableStateOf("") }
+  var ageToText by remember { mutableStateOf("") }
```
arthur-mrgt, line 101: ageFromInput & ageToInput would be a more consistent naming

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,323 @@
...
+  val context = LocalContext.current
+
+  val textFieldHeightFactor = 0.08f
+  val textFieldWidthFactor = 0.2f
+  val buttonHeightFactor = 0.07f
+  val buttonWidthFactor = 0.3f
+  val tagSelectorHeightFactor = 0.25f
+  val tagSelectorWidthFactor = 0.8f
+  val tagSelectorTextSizeFactor = 0.1f
+  val titleFontSizeFactor = 0.03f
```
arthur-mrgt, line 68: Magic numbers like theses should be constants.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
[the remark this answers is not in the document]
arthur-mrgt: I think it is indeed a good idea to combine these two functions and make one with this signature:

fun validateAndUpdateAge(input: String, isFrom: Boolean) { /*...*/ }

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,323 @@
...
+  var womanSelected by remember { mutableStateOf(false) }
+  var otherSelected by remember { mutableStateOf(false) }
+  var ageFrom by remember { mutableStateOf<Int?>(null) }
+  var ageTo by remember { mutableStateOf<Int?>(null) }
+
+  // test values
+  var selectedTags = remember {
+    mutableStateOf<List<Pair<String, Color>>>(listOf(Pair("salsa", Color.Red)))
+  }
+  val availableTags = remember {
```
arthur-mrgt, line 86: Store this outside or use rememberSaveable if it can change.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -0,0 +1,323 @@
...
+              "Age Range",
+              fontSize = (screenHeight.value * titleFontSizeFactor).sp,
+              modifier = Modifier.padding(vertical = 8.dp).testTag("AgeRangeTitle"))
+          val textFieldHeight = screenHeight * textFieldHeightFactor
+          val textFieldWidth = screenWidth * textFieldWidthFactor
+
+          Row(
+              modifier = Modifier.fillMaxWidth().padding(horizontal = 16.dp),
+              horizontalArrangement = Arrangement.Center) {
+                OutlinedTextField(
```
arthur-mrgt, line 211: Maybe try to modularize better by extracting this composable. This would make the code more readable and easier to maintain.

arthur-mrgt (changes requested):
The FilterScreen composable effectively implements the required functionality, but it could benefit significantly from improved modularity. Here are key points for enhancement:

1- Extract Composables: Break down FilterScreen into smaller, reusable composables, such as AgeInputField, GenderSelection, and TagSelector. This enhances clarity and reusability.

2- Single Responsibility Principle: Ensure each composable has a single responsibility, simplifying testing and reducing complexity. Smaller, well-named composables improve code readability, making it easier for others to understand your intentions.

3- State Management: Group related state variables into a data class to streamline state handling.

→ Modular components can be tested individually, leading to a more robust codebase.

arthur-mrgt (approved):
Working code with excellent coverage, approved for now, but will need better modularization in the future (see my previous comment).

BotondAKovacs:
@arthur-mrgt is absolutely correct I should modularise better. I'll update Filter.kt in the future with this in mind. However it works and we can work with it for now.

PR #26, opened 2024-10-16
feature/add-profiles-viewmodel

arthur-mrgt:
# Profiles ViewModel

## Overview
This pull request introduces the `ProfilesViewModel` for managing user profiles and the corresponding unit tests to ensure its functionality. 

### Features Added
- **ProfilesViewModel**: 
  - Handles profile fetching, filtering, and management.
  - Implements state management for profiles, selected profile, loading state, and error handling.
  - Methods for:
    - Fetching filtered profiles within a radius.
    - Adding, updating, retrieving, and deleting profiles by UID.
  - Integrated with Firebase Firestore for data operations.

### Tests Added
- **ProfilesViewModelTest**: 
  - Unit tests for `ProfilesViewModel` using Mockito.
  - Tests for:
    - Fetching filtered profiles in a radius.
    - Adding, updating, retrieving, and deleting profiles.
    - Error handling to ensure proper state management on failures.
  - Verification of interaction with `ProfilesRepository` to ensure correct method calls and parameters.

## Additional Notes
- All tests are designed to mock the `ProfilesRepository` to isolate the `ProfilesViewModel` logic.

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -1,0 +1,154 @@
+package com.github.se.icebreakrr.model.profile
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import com.google.firebase.Firebase
+import com.google.firebase.firestore.GeoPoint
+import com.google.firebase.firestore.firestore
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+open class ProfilesViewModel(private val repository: ProfilesRepository) : ViewModel() {
+
+    private val _profiles = MutableStateFlow<List<Profile>>(emptyList())
+    val profiles: StateFlow<List<Profile>> = _profiles
+
+    private val _selectedProfile = MutableStateFlow<Profile?>(null)
+    val selectedProfile: StateFlow<Profile?> = _selectedProfile
+
+    private val _loading = MutableStateFlow(false)
+    val loading: StateFlow<Boolean> = _loading
+
+    private val _error = MutableStateFlow<Exception?>(null)
+    val error: StateFlow<Exception?> = _error
+
+    init {
+        repository.init { //TODO getProfilesInRadius}
+        }
+    }
+
+    companion object {
+        val Factory: ViewModelProvider.Factory =
+            object : ViewModelProvider.Factory {
+                @Suppress("UNCHECKED_CAST")
+                override fun <T : ViewModel> create(modelClass: Class<T>): T {
+                    return ProfilesViewModel(ProfilesRepositoryFirestore(Firebase.firestore)) as T
+                }
+            }
+    }
+
+    /**
+     * Fetches profiles within a given radius around a center point with optional filters.
+     *
+     * @param center The center location as a GeoPoint.
+     * @param radiusInMeters The radius around the center in meters to search for profiles.
+     * @param gender Optional filter for gender.
+     * @param ageRange Optional filter for age range.
+     * @param tags Optional filter for specific tags.
+     */
+    fun getFilteredProfilesInRadius(
+        center: GeoPoint,
+        radiusInMeters: Double,
+        gender: Gender? = null,
+        ageRange: IntRange? = null,
+        tags: List<String>? = null
+    ) {
+        _loading.value = true
+        repository.getProfilesInRadius(
+            center = center,
+            radiusInMeters = radiusInMeters,
+            onSuccess = { profileList ->
+                val filteredProfiles = profileList.filter { profile ->
+
+                    // Filter by gender if specified
+                    (gender == null || profile.gender == gender) &&
+
+                            // Filter by age range if specified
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -0,0 +1,148 @@
...
+   * @param radiusInMeters The radius around the center in meters to search for profiles.
+   * @param gender Optional filter for gender.
+   * @param ageRange Optional filter for age range.
+   * @param tags Optional filter for specific tags.
+   */
+  fun getFilteredProfilesInRadius(
+      center: GeoPoint,
+      radiusInMeters: Double,
+      gender: Gender? = null,
+      ageRange: IntRange? = null,
```
louiscrc, line 53: It makes sense that gender and tags are optionnals, but why the ageRange ? From the UI, I think ageRange always has a value

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -0,0 +1,148 @@
...
+   * @param center The center location as a GeoPoint.
+   * @param radiusInMeters The radius around the center in meters to search for profiles.
+   * @param gender Optional filter for gender.
+   * @param ageRange Optional filter for age range.
+   * @param tags Optional filter for specific tags.
+   */
+  fun getFilteredProfilesInRadius(
+      center: GeoPoint,
+      radiusInMeters: Double,
+      gender: Gender? = null,
```
louiscrc, line 52: You can have several genders preferences, so I think this should be a List<Gender> rather than a single Gender

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -0,0 +1,148 @@
...
+              profileList.filter { profile ->
+
+                // Filter by gender if specified
+                (gender == null || profile.gender == gender) &&
+
+                    // Filter by age range if specified
+                    (ageRange == null || profile.calculateAge() in ageRange) &&
+
+                    // Filter by tags if specified
+                    (tags == null || profile.tags.containsAll(tags))
```
louiscrc, line 71: I think this should be containsAny rather than contains all.
If you set your preferences to football and pasta, it means that you can only see people who have set both football and pasta in their profile. 
The point would be more to see people who share one of your interests rather than all your interests.

louiscrc: But this is temporary because Samuel is currently implementing, in his TagsViewModel, a filter such that when you are interested by football, which is in the category sports, you can still see the profiles of people interested by sports globally.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -0,0 +1,148 @@
...
+   */
+  fun deleteProfileByUid(uid: String) {
+    _loading.value = true
+    repository.deleteProfileByUid(
+        uid,
+        onSuccess = { _loading.value = false },
+        onFailure = { exception ->
+          _error.value = exception
+          _loading.value = false
+        })
```
louiscrc, line 146: Be careful about code duplication. onSucess and onFailure are always the same and use class variables. You could make two independant functions handleSuccess and handleFailure, and use them for each of your functions

louiscrc (changes requested):
Your test coverage is excellent.

Please take a look at the different comments I made for the ViewModel, especially for code duplication.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
arthur-mrgt: Even if this is indeed the case, it is always better to have a default value. If the filter is null it does not filter anything, which is the case when the user leaves the default values

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
arthur-mrgt: Good catch, I will correct this

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
arthur-mrgt: I don't really agree with this remark. The classic behavior expected by the user with this type of filter is to correspond exactly to the request. For example if the user puts the tag "football" and "pizza" he expects to see only profiles which contain these two tags (in addition to others).

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
arthur-mrgt: I did an handleError function to improve code reusability.

louiscrc (approved):
You implemented very well my remarks.
Thank you for your work !

PR #27, opened 2024-10-16
Feature/tagsBackend viewModel

SamSteu:
**General Description**
In this PR I created the `ViewModel ` for the tags with functions useful to do all sorts of operations on the tags.

in `TagsViewModel` :
- `Factory` method that creates the `TagsViewModel` object
- `init` method called at the creation of the object. All it does is store the tags found in the database into a private variable `allTag` by calling the `init` from `TagsRepository`
- `private tagToCategory` utility function that takes a `String` (tag) and returns the `CategoryString enum` from where the tag correspond. Example : `tagToCategory("Basketball") = CategoryString.Sport`
- `tagToColor` takes a `String` (tag) in parameter and returns the `Color` in which it must be showed
- `setQuery` is a function that must be called when we search for tags (EditProfile and Filter). It modifies the observable variable `tagsSuggestion` that represents items in the `DropdownMenu`
- `applyFilters` needs to be called when we save the filters. It will modify the observable variable `filteringTags` that represents all tags that you can see in the Around You page
- `addFilter` needs to be called when we select a new tag in the Filter page and `removeFilter` needs to be called when we click on a tag to remove it in the Filter page. This will modify the observable variable `filteringTags` which represents the tags that will filter who you can see in the Around You page
- `leaveUI` needs to be called whenever you quit the EditProfile page or the Filter page. Otherwise there will be conflicts between them. It resets the observable variables `query` and `tagsSuggestions `

in TagsRepository : 
- init method that checks if there is a current user on firebase

in `MainActivity` : 
- I added the instantiation of the `TagsViewModel` and put them into the needed composables (`AroundYou` and `NotificationScreen` for the moment)

in `TagsViewModelTest` I added tests for the `TagsViewModel` class


app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -24,24 +24,14 @@ import com.github.se.icebreakrr.ui.sections.FilterScreen
...
     super.onCreate(savedInstanceState)
-
-    // Initialize Firebase Auth
-    FirebaseApp.initializeApp(this)
     auth = FirebaseAuth.getInstance()
-    auth.currentUser?.let {
-      // Sign out the user if they are already signed in
-      // This is useful for testing purposes
-      auth.signOut()
-    }
```
MaxCastelCast, line 44: I think this piece of code was useful for authentification

gradle/libs.versions.toml
```diff
@@ -1,67 +1,68 @@
...
-espressoCore = "3.5.1"
-appcompat = "1.6.1"
-material = "1.11.0"
-composeBom = "2024.02.02"
-composeActivity = "1.8.2"
-composeViewModel = "2.7.0"
-lifecycleRuntimeKtx = "2.7.0"
-kaspresso = "1.5.5"
-mockitoAndroid = "5.13.0"
+#Google
```
MaxCastelCast, line 2: Nice to reorganisse the version controler, very good !

MaxCastelCast (approved):
Looks good to me

PR #31, opened 2024-10-17
fix: add distribution version and test-results uploading

louiscrc:
Added few fixes.

There has to be a PR per change because this can be tested only within the main.


MaxCastelCast (approved):
LGTM

PR #32, opened 2024-10-17
fix: change JDK version 11 -> 17

louiscrc:


Kapestian (approved):
lgtm

PR #33, opened 2024-10-17
fix: add env for the tests

louiscrc:


MaxCastelCast (approved):
LGTM

PR #34, opened 2024-10-17
fix: copy the kestore and google-services from secrets

louiscrc:
> [!NOTE]
> I want the readers to read it carefully as it contains many important docs.


MaxCastelCast (approved):
LGTM

PR #35, opened 2024-10-17
fix: add env for the build

louiscrc:
(should work this time I promise)


MaxCastelCast (approved):
lgtm

PR #36, opened 2024-10-17
Feature/profile screen

JanStaszewiczEPFL:
### The profile screen

summary : Added the profile screen and merged it with other features

**profile**

- As for now added a false user with predefined data with two sections : the `header` and the `info`. In the header there is the profile picture, the username, the edit button and the go back button. In the info section there is the catch phrase, the tags and the description.
- As for now the `settings` screen is replaced by the `profile screen`. When you click on a `profile card` in the `around you` screen it redirects to the fake profile.

profileScreen

**testing**

I reworked the navigation testing to take into account the new user flow for the settings `screen` and the `profile cards`

### DIFF
--- a/app/src/androidTest/java/com/github/se/icebreakrr/ui/profile/ProfileTest.kt
+++ b/app/src/androidTest/java/com/github/se/icebreakrr/ui/profile/ProfileTest.kt
@@ -1,0 +1,103 @@
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.test.*
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.se.icebreakrr.ui.profile.ProfileView
+import com.github.se.icebreakrr.ui.tags.RowOfTags
+import com.github.se.icebreakrr.ui.navigation.NavigationActions
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
+
+@RunWith(AndroidJUnit4::class)
+class ProfileViewTest {
+
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    private lateinit var navigationActionsMock: NavigationActions
+
+    @Before
+    fun setUp() {
+        navigationActionsMock = mock()
+    }
+
+    @Test
+    fun testProfileHeaderDisplaysCorrectly() {
+        composeTestRule.setContent {
+            ProfileView(navigationActions = navigationActionsMock)
+        }
+
+        // Verify that the profile header and its elements are displayed
+        composeTestRule.onNodeWithTag("profileHeader").assertIsDisplayed()
+        composeTestRule.onNodeWithTag("profilePicture").assertIsDisplayed()
+        composeTestRule.onNodeWithTag("goBackButton").assertIsDisplayed()
+        composeTestRule.onNodeWithTag("editButton").assertIsDisplayed()
+        composeTestRule.onNodeWithTag("username").assertTextEquals("Jean Michelle")
+    }
+
+    @Test
+    fun testGoBackButtonFunctionality() {
+        composeTestRule.setContent {
+            ProfileView(navigationActions = navigationActionsMock)
+        }
+
+        composeTestRule.onNodeWithTag("goBackButton").assertIsDisplayed()
+        composeTestRule.onNodeWithTag("goBackButton").performClick()
+
+        verify(navigationActionsMock).goBack()
+    }
+
+    @Test
+    fun testEditButtonFunctionality() {
+        composeTestRule.setContent {
+            ProfileView(navigationActions = navigationActionsMock)
+        }
+
+        composeTestRule.onNodeWithTag("editButton").assertIsDisplayed()
+        composeTestRule.onNodeWithTag("editButton").performClick()
+
+        //Add edit profile navigation logic
+    }
+
+    @Test
+    fun testInfoSectionDisplaysCorrectly() {
+        composeTestRule.setContent {
+            ProfileView(navigationActions = navigationActionsMock)
+        }
+
+        // Verify that the info section is displayed
+        composeTestRule.onNodeWithTag("infoSection").assertIsDisplayed()
+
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/se/icebreakrr/ui/sections/AroundYou.kt
```diff
@@ -68,9 +68,7 @@ fun AroundYouScreen(
                 items(profiles.value.size) { index ->
                   ProfileCard(
                       profile = profiles.value[index],
-                      onclick = {
-                        Toast.makeText(context, "Profile clicked", Toast.LENGTH_SHORT).show()
-                      })
+                      onclick = { navigationActions.navigateTo("Settings Screen") })
```
Kapestian, line 71: question: Why `"Settings Screen"` instead of a route like `"Profile View"` ? But most importantly, why does it work ???

app/src/androidTest/java/com/github/se/icebreakrr/ui/profile/ProfileTest.kt
```diff
@@ -0,0 +1,99 @@
...
+  @Test
+  fun testProfileDescriptionDisplaysCorrectly() {
+    composeTestRule.setContent { ProfileView(navigationActions = navigationActionsMock) }
+
+    // Verify that the profile description is displayed correctly
+    composeTestRule
+        .onNodeWithTag("profileDescription")
+        .assertTextContains(
+            """
+                Hey there! I'm John Do, a bit of a tech geek who loves exploring all things digital.
```
Kapestian, line 90: `Hey there! I'm John Do`
Nitpick: here is John Do but under the picture it is Jean-Michel

app/src/main/res/drawable/turtle.gif
Kapestian, line 1: I just love it

Kapestian (commented):
Just one nitpick (kind of immersion breaking). Please make this change and then you're done. Nice job mate!

app/src/main/java/com/github/se/icebreakrr/ui/sections/AroundYou.kt
JanStaszewiczEPFL: Because I place the ProfileView inside the settings screen temporarily

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileView.kt
```diff
@@ -0,0 +1,244 @@
...
+ * Displays the main profile view with the profile header and information section. The profile
+ * header shows the profile image, back button, and user name. The information section contains a
+ * catchphrase, tags, and description.
+ *
+ * @param navigationActions Actions to navigate between screens.
+ */
+@Composable
+fun ProfileView(navigationActions: NavigationActions) {
+
+  // Predefined user for UI testing
```
arthur-mrgt, line 54: Perhaps we could create a "Profile" object and then use it throughout the rest of the code. This would align better with future logic and simplify the final implementation.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileView.kt
```diff
@@ -0,0 +1,244 @@
...
+        Column(
+            modifier = Modifier.fillMaxWidth().padding(paddingValues),
+            horizontalAlignment = Alignment.CenterHorizontally) {
+              // 2 sections one for the profile image with overlay and
+              // one for the information section
+              ProfileHeader(navigationActions)
+              InfoSection(
+                  catchPhrase = catchphrase,
+                  listOfTags = userTags,
+                  description =
```
arthur-mrgt, line 87: Here, use the Profile object, such as mockProfile.catchphrase, mockProfile.catchPhrase, etc.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileView.kt
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/**
+ * Displays a section with a row of tags, each associated with a color.
+ *
+ * @param listOfTags A list of tags, each paired with a color.
+ */
+@Composable
+fun TagsSection(listOfTags: List<Pair<String, Color>>) {
+  Box(modifier = Modifier.fillMaxWidth().height(80.dp).testTag("tagSection")) {
```
arthur-mrgt, line 225: Don't use magic values, define them as constants for better maintainability: 

private const val TAG_SECTION_HEIGHT = 80.dp

app/src/androidTest/java/com/github/se/icebreakrr/ui/profile/ProfileTest.kt
JanStaszewiczEPFL: True, I'll fix that!

app/src/androidTest/java/com/github/se/icebreakrr/ui/profile/ProfileTest.kt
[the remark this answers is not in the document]
JanStaszewiczEPFL: I agree!

arthur-mrgt (changes requested):
# Code Review: ProfileView

Overall, excellent work done on the `ProfileView` component. The code is well modularized and very clear, making it easy to understand and maintain. The documentation is well-written and provides valuable context for each function. Additionally, the coverage is excellent, ensuring that we have robust testing in place.

For tomorrow's demo, this implementation works as expected. However, moving forward, we will need to add the logic to display either the currently logged-in user's profile or a profile they wish to view. To achieve this, we need to determine the origin of the user's action (did they click the "profile" button or a profile from the "Around You" page?). 

Once we establish this context, we will retrieve the UID of the profile to display using either `FirebaseAuth.getInstance().currentUser` or `selectedProfile.uid`. If necessary, we can fetch the profile data using `getProfileByUid`. Finally, we need to ensure that the retrieved profile object is displayed correctly in the UI.

This last point could be implemented right away by creating a Profile object (named mockedProfile) for the demo, which can then be used throughout the rest of the code.

app/src/main/res/drawable/turtle.gif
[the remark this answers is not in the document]
JanStaszewiczEPFL: Ahahah I agree!

arthur-mrgt (approved):

PR #37, opened 2024-10-17
Feature/notification screen

MaxCastelCast:
For Botond

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
@@ -2,7 +2,6 @@ package com.github.se.icebreakrr.ui.sections
 
 import android.annotation.SuppressLint
 import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
```diff
@@ -20,20 +32,43 @@ import com.github.se.icebreakrr.ui.navigation.NavigationActions
  */
 @SuppressLint("UnusedMaterial3ScaffoldPaddingParameter")
```
SamSteu, line 33: I think that you should use the internal paddings of the Scaffold so that the screen adapts better to the screen size and the content doesn't overlap with the top and bottom bar.

SamSteu (commented):
If the code looks visualy well (according to my comment), we can merge it so that we can use it for the demo tomorrow and add in the scrumboard a task to make the screen more adaptive.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
MaxCastelCast: Yes indeed, you are rigth, but we could let that for the next sprint

SamSteu (approved):

PR #41, opened 2024-10-17
Revert "Feature/notification screen"

Kapestian:
Reverts Swent-team-6/icebreakrr#37 as there was actually some changes to make


MaxCastelCast (approved):
I agree with the revert, need to add the ProfileView screen in the navigations

PR #42, opened 2024-10-17
Restore "Feature/notification screen"

Kapestian:
the revert Didn't work well, so we are restoring the PR and I'll create another PR to fix the issues pointed by the missed review comments on the original


MaxCastelCast (approved):
lgtm

PR #43, opened 2024-10-17
Feature/settings screen UI

BotondAKovacs:
Summary:
Create the UI of the Settings screen, adapting the navigation so that going to the profile goes through the settings page and finally creating a variant of the profile card so that the settings screen can have its own variant.

### DIFF
--- a/app/src/androidTest/java/com/github/se/icebreakrr/ui/sections/SettingsTest.kt
+++ b/app/src/androidTest/java/com/github/se/icebreakrr/ui/sections/SettingsTest.kt
@@ -1,0 +1,85 @@
+package com.github.se.icebreakrr.ui.sections
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertIsOff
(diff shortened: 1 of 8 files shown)


app/src/androidTest/java/com/github/se/icebreakrr/ui/sections/SettingsTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+  @Test
+  fun testNavigationActionsOnProfileCardClick() {
+    composeTestRule.setContent { SettingsScreen(navigationActionsMock) }
+
+    // Click the profile card
+    composeTestRule.onNodeWithTag("profileCard").performClick()
+
+    // Verify that the correct navigation action is triggered
+    verify(navigationActionsMock).navigateTo(Screen.PROFILE)
```
JanStaszewiczEPFL, line 69: You could maybe check here if the profile screen is indeed displayed with assertIsDisplayed().

Kapestian (approved):
LGTM

PR #46, opened 2024-10-17
test : add end to end tests

MaxCastelCast:
This end to end test shows all the systems that should work form M1 : The navigations between screens and the UI logic.


BotondAKovacs (approved):
LGTM

PR #50, opened 2024-10-31
Feature/message UI

SamSteu:
This PR adds the UI to request a meeting with someone. It is not yet connected to the backend. 

in `TopLevelDestination` : 
- I added a new screen `AroundYouProfile` which is like the `ProfileView` but with a message icon instead of an edit Icon. There is a lot of code from `ProfileView` that I used too soo I moved that code in the shared package
in `MainActivity` : 
- I added the new screen to the navigation graph
I added a new file `Request` : 
- UI that represents the dialog popup to send a request
I added a new file `AroundYouProfile` : 
- When you click on a profile in the `AroundYou` screen, it is the UI of this profile
In `ProfileView` : 
- Utility functions that I use in the `AroundYouProfile` moved to the shared package
in `AroundYou` : 
- When you click on a profile it doesn't go anymore on the `ProfileView` but on `AroundYouProfile`
I added a new file `DisplayProfile` :  
- This is the shared file used by `ProfileView` and `AroundYouProfile`. I changed some parameters so that the functions are usable by the two

test :
- I modified the `M1_test` end-to-end to fit the changes I made.
- I added a new file `AroundYouProfileTest` that tests the UI of the two images below
- I modified `AroundYouTest` so that when you click on a card it navigates to the screen `AROUND_YOU_PROFILE` instead of `PROFILE`

When you click on a profile in `AroundYou` :
image

When you want to send a request to this profile : 
image


JanStaszewiczEPFL:
We discussed the option of using a single profile screen that would dynamically adapt to different user data inputs. However, keeping AroundYouProfile as a separate screen works well too. It might be helpful to rename it to reflect and keep in mind that it’s a general profile view for other users, not limited to being accessed only from the "Around You" screen.

JanStaszewiczEPFL (changes requested):
We discussed the option of using a single profile screen that would dynamically adapt to different user data inputs. However, keeping AroundYouProfile as a separate screen works well too. It might be helpful to rename it to reflect and keep in mind that it’s a general profile view for other users, not limited to being accessed only from the "Around You" screen.

app/src/main/java/com/github/se/icebreakrr/ui/message/Request.kt
```diff
@@ -0,0 +1,106 @@
...
+                    horizontalAlignment = Alignment.Start,
+                ) {
+                  Text(
+                      text = "Meet this person",
+                      style =
+                          TextStyle(
+                              fontSize = 24.sp,
+                              lineHeight = 32.sp,
+                              fontWeight = FontWeight(400),
+                              color = Color(0xFF1D1B20),
```
MaxCastelCast, line 62: The color is hardcoded, should use color constants in the resource folder or add your colors there, could be done later tho

JanStaszewiczEPFL (approved):
LGTM

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -90,6 +91,7 @@ fun IcebreakrrApp() {
       composable(Screen.AROUND_YOU) {
         AroundYouScreen(navigationActions, profileViewModel, tagsViewModel)
       }
+      composable(Screen.OTHER_PROFILE_VIEW) { OtherProfileView(navigationActions) }
```
MaxCastelCast, line 94: Good idea i think, makes things clearer to differenciate your own profile to the other profiles

MaxCastelCast (approved):
There are some colors and lengths hardcoded in the code but the rest looks good to me

app/src/main/java/com/github/se/icebreakrr/ui/message/Request.kt
SamSteu: Yep I will change that in my next PR since it will also be on the messaging system!

PR #52, opened 2024-10-31
Feature/filter view model and linking

BotondAKovacs:
Connect Filter UI with Backend Logic

Overview
This PR implements the backend logic for the existing filter UI, by linking the backends and saving user input

Changes
- Added `FilterViewModel` to save user inputs (gender, age range, tags)
- Connected the filter UI to the view model
- The around you page profiles are correctly filtered with the filter page filters

Testing
Added unit tests covering the view model

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -77,13 +77,15 @@ open class ProfilesViewModel(private val repository: ProfilesRepository) : ViewM
               profileList.filter { profile ->
 
                 // Filter by genders if specified
-                (genders == null || profile.gender in genders) &&
+                (genders == null || profile.gender in genders || genders.isEmpty()) &&
 
                     // Filter by age range if specified
                     (ageRange == null || profile.calculateAge() in ageRange) &&
 
                     // Filter by tags if specified
-                    (tags == null || profile.tags.containsAll(tags))
+                    (tags == null ||
+                        profile.tags.any { it.lowercase() in tags.map { it.lowercase() } } ||
+                        tags.isEmpty())
               }
           _profiles.value = profileList
           _filteredProfiles.value = filteredProfiles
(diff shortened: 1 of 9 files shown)


MaxCastelCast (approved):
Looks good to me

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -77,13 +77,15 @@ open class ProfilesViewModel(private val repository: ProfilesRepository) : ViewM
...
                 // Filter by genders if specified
-                (genders == null || profile.gender in genders) &&
+                (genders == null || profile.gender in genders || genders.isEmpty()) &&
 
                     // Filter by age range if specified
                     (ageRange == null || profile.calculateAge() in ageRange) &&
 
                     // Filter by tags if specified
-                    (tags == null || profile.tags.containsAll(tags))
+                    (tags == null ||
```
arthur-mrgt, line 86: It seems more logical to me to display the profiles that have all the tags that the user requested.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
```diff
@@ -266,8 +286,34 @@ fun FilterScreen(
...
+                    }
+                    ageFrom == null && ageTo != null -> {
+                      filterViewModel.setAgeRange(0..ageTo!!)
+                    }
+                    else -> {
+                      filterViewModel.setAgeRange(null)
+                    }
+                  }
+
                   navigationActions.goBack()
```
arthur-mrgt, line 317: You need to call getFilteredProfilesInRadius() with the updated filters to refresh the filteredProfiles list and recompose the UI. For now, only manual updates through the pull-to-refresh process will refresh the list, which is not the expected behavior.

arthur-mrgt (changes requested):
LGTM! You just need to add a call to getFilteredProfilesInRadius() when the filter page’s validation button is pressed.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
BotondAKovacs: We discussed it a bit on the private discord, the problem is that the tagviewmodel was developed with "or" in the devs' mind, not an "and" so it would be very complicated for me to change that in the scope of this PR. Also as a team we haven't agreed on what's best as there are good points for both ideas. We'll see in the future if we change this

app/src/main/java/com/github/se/icebreakrr/ui/sections/Filter.kt
BotondAKovacs: You're right, I'll fix it.

arthur-mrgt (approved):
Perfect. In future PRs, test coverage will need to be improved.

arthur-mrgt (approved):
LGTM

PR #55, opened 2024-11-02
Feature/message UI

SamSteu:
**1. Modify UI for your profile and others profile**
in `OtherProfileView` : 
- I changed the hardcoded list of tags to have the correct colors
in `DisplayProfile` : 
- I changed the order in which elements are displayed. Now there is the catchphrase, then the description and then the tags. 
- Every field (except the catch phrase) has a title in _IceBreakrrBlue_ so that it fits better the figma
- Changed the appearance of the edit and send message button
in `OtherProfileView` : 
- changed the hardcoded value of the tags to hardcoded values with the right color

image

image

**2. Modifiy UI for th tag selector** 
in `TagView` : 
- make put a max height to the dropdown menu so that it doesn't go over the keyboard
- if there are less than 5 elements in the dropdown menu, make the height of the dropdown menu adaptable to the number of items
- when you search a tag that doesn't exist, make the dropdown menu not expanded anymore

https://github.com/user-attachments/assets/923499fd-7e92-4c96-8ff9-d759bfbc7553

**Other files :** 
in TagsViewModel I changed the `.lowercase(Local.ROOT)` to `.lowercase()`


app/src/main/java/com/github/se/icebreakrr/ui/sections/shared/DisplayProfile.kt
```diff
@@ -50,6 +53,21 @@ import com.github.se.icebreakrr.ui.tags.TagStyle
...
+val catchPhraseLineHeight = 28.sp
+val catchPhraseWeight = 400
+val descriptionFontSize = 16.sp
+val descriptionLineHeight = 24.sp
+val descriptionFontWeight = 500
+val descriptionLetterSpacing = 0.15.sp
+val tagHeight = 40
+val requestButtonSize = 55.dp
+val requestButtonElevation = 8.dp
+
```
MaxCastelCast, line 70: Nice ! All the constants are now stored in the shared folder, that makes hte code clearer and easier to maintain

app/src/main/java/com/github/se/icebreakrr/ui/tags/TagView.kt
```diff
@@ -35,6 +36,9 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import androidx.compose.ui.window.PopupProperties
 
+val dropDownMenuTagHeight = 60
+val dropDownMenuMaxNumberTag = 5
+
```
MaxCastelCast, line 41: This is a nice debug, we needed to make the DropDownMenu smaller, it taked all the screen before, now it is cleaner

MaxCastelCast (approved):
Looks good to me !

PR #57, opened 2024-11-05
Feature/profile edit

louiscrc:
In this pull request, I did quite few changes on the profile editing functionality, essentially by linking the ProfileEditing to the backend, and I associated test cases for the UI and the application of the ViewModel. I also implemented the reactivity to the screen size.

#### Profile Editing Screen Enhancements:
* Added `ProfilesViewModel` and `TagsViewModel` dependencies to `ProfileEditingScreen` to manage profile data and tags.
* Implemented dynamic UI elements such as profile picture size and text field heights based on screen dimensions.
* Added loading state handling and profile data fetching using Firebase authentication.
* Introduced confirmation dialogs for unsaved changes when navigating away from the profile editing screen. [[1]](diffhunk://#diff-1b70eef1aa85d92e1fef4471d75b6c9af8782cce50565ff0e3e58ea794ba77a6L102-R113) [[2]](diffhunk://#diff-1b70eef1aa85d92e1fef4471d75b6c9af8782cce50565ff0e3e58ea794ba77a6L158-R225)

#### Test Case Updates:
* Updated `ProfileEditingScreenTest` to include new ViewModel dependencies and additional test cases for loading state, top bar elements, and profile content validation.
* Refactored existing test cases to accommodate new UI elements and interactions, such as checking for loading indicators and verifying profile data.

#### Navigation and Layout Adjustments:
* Modified `MainActivity` to pass the new `ProfilesViewModel` to the `ProfileEditingScreen` composable.
* Refactored layout code in `ProfileEdit.kt` to improve readability and maintainability, including the use of `LaunchedEffect` for side effects.


app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
```diff
@@ -1,215 +1,229 @@
...
+                        TextButton(
+                            onClick = {
+                              showDialogConfirm = false
+                              updateProfile()
+                              tagsViewModel.leaveUI()
+                              navigationActions.goBack()
+                            }) {
+                              Text("Save")
+                            }
+                      },
```
Kapestian, line 220: Design: Even if having a dialog to confirm data update, it doesn't feel right when using the app. I think adding a prompt to confirm the changes adds too much friction to the edit process and is not required, as a user can always re-edit it's profile if something is missing or wrong.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
```diff
@@ -1,215 +1,229 @@
...
+                      confirmButton = {
+                        TextButton(
+                            onClick = {
+                              showDialogBack = false
+                              tagsViewModel.leaveUI()
+                              navigationActions.goBack()
+                            }) {
+                              Text("Discard changes")
+                            }
+                      },
```
Kapestian, line 200: nitpick: could you add a boolean variable or something to check if something was changed and only then show the prompt when going back? Otherwise the prompt adds friction for nothing.

Kapestian (changes requested):
Your Changes are very good, I am really happy to see our idea come to life. The Only request that I have is to change the behavior of the confirmation popup so it reduces friction (we want the navigation to be as smooth as possible). The details are explained in my comments
When this change will be done (I think it will be quickly resolved), get back to me and I'll approve.

Great work !

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
```diff
@@ -1,215 +1,229 @@
...
-          name = "",
-          gender = Gender.OTHER,
-          birthDate = Timestamp.now(),
-          catchPhrase = "",
-          description = "",
-          tags = listOf(),
-          profilePictureUrl = "")
+  LaunchedEffect(Unit) {
+    Firebase.auth.currentUser?.let { profilesViewModel.getProfileByUid(it.uid) }
+  }
```
Kapestian, line 54: Suggestion: now that I think of it, we could directly ad that in the settings screen as it is needed for the setting screen card, then your own profile's preview and only then edit profile. By doing this `getProfileByUid` only once we could reduce the number of network call, load every page (after setting) faster and reduce the probability of net error

louiscrc: I agree with you, thus there is already 38 implementations of profilesViewModel.getProfileByUid(it.uid) especially in the tests of the viewmodel and the firebase, thus I'd let the author (Arthur) do it.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
louiscrc: I actually thought about doing it but I think something blocked me. I will try again.

Kapestian (approved):
Perfect, thank you for the changes!

Kapestian (approved):

PR #64, opened 2024-11-08
Feature/upload pp

Kapestian:
# Upload profile pictures
This pull request introduces significant updates to the profile pictures display and selection. The primary changes include a profile picture selector that picks photos right from the gallery without having to use permissions. It also replaces pictures placeholders with actual profile pictures, which are stored in Firebase storage.

> [!note]  
**Revievers**: Always check at the bottom of the description as you will find a list of things that need to be checked first (doubts I had, decisions I have taken that need to be double-checked, ...)

ezgif-3-7d3cb5cb54 image 

## Changes
- **ProfilesViewModel.kt**
  - Added functions to process image for upload.
  - A new StateFlow that allows the UI to access a temporary processed file for display before upload.
  - Fix mistakes in last week's PR

- **ProfilePictureSelector.kt**
  - Added a new composable function `ProfilePictureSelector` to allow users to select and display profile pictures.
  - Implemented image picker functionality to choose images from the gallery.

- **ProfileEdit.kt**
  - Introduced the new `ProfilePictureSelector` composable for profile picture selection and display.

- **ProfileCard.kt**
  - Replaced empty circles with actual profile pictures

To view more details, please refer to the [pull request files](https://github.com/Swent-team-6/icebreakrr/pull/64/files).

## Review these first!
Here is a list of details I am not sure about
* I am unsure of the efficiency of my tests as it uses a lot of mocking and I don't know how much is too much.
* [SOLVED] When the picture is selected in ProfileEdit, taping the discard button to go back to the profile preview won't discard the profile picture. This difference with @louiscrc editing logic is due to the fact that doing so will lead to bad separation of layers (UI and Logic) and callback hell. This is not a big deal as it still makes sense that image and text may have different logic
* I am not sure about the error handling. Using callbacks and @arthur-mrgt `handleError` function is confusing. Maybe I am doing things wrong. The only thing I know is that the errors I could get in my function are non-terminating, so It must work anyway.
* Launcher, activities, contracts... I am completely lost with that. If someone is comfortable with that, please double-check my code (and if you have good resources to learn that, I'll take a look at them)
* [SOLVED]  I am searching for a way to use my processed square jpg as a placeholder to the profile picture selector while the actual image is being uploaded. If you have an idea how to do that dont hesitate to reach out

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -203,23 +203,32 @@ open class ProfilesViewModel(
   }
 
     fun imageUriToJpgByteArray(context: Context, imageUri: Uri, quality: Int = 100): ByteArray? {
-        return try {
-            // open an InputStream from the URI
-            val inputStream: InputStream? = context.contentResolver.openInputStream(imageUri)
+    return try {
+        // open an InputStream from the URI
+        val inputStream: InputStream? = context.contentResolver.openInputStream(imageUri)
 
-            // decode InputStream to Bitmap
-            val bitmap = BitmapFactory.decodeStream(inputStream)
-            inputStream?.close() // close the InputStream after decoding
+        // decode InputStream to Bitmap
+        val bitmap = BitmapFactory.decodeStream(inputStream)
+        inputStream?.close() // close the InputStream after decoding
 
-            //compress Bitmap to JPEG format and store in ByteArrayOutputStream
-            val byteArrayOutputStream = ByteArrayOutputStream()
-            bitmap?.compress(Bitmap.CompressFormat.JPEG, quality, byteArrayOutputStream)
+        // todo: delegate image cropping to another function
+        // calculate the dimensions for the square crop
+        val dimension = minOf(bitmap.width, bitmap.height)
+        val xOffset = (bitmap.width - dimension) / 2
+        val yOffset = (bitmap.height - dimension) / 2
 
-            // return ByteArray of compressed JPEG image (representing .jpg)
-            byteArrayOutputStream.toByteArray()
-        } catch (e: Exception) {
-            e.printStackTrace()
-            null
-        }
+        // crop the Bitmap to a square at the center
+        val croppedBitmap = Bitmap.createBitmap(bitmap, xOffset, yOffset, dimension, dimension)
+
+        // compress Bitmap to JPEG format and store in ByteArrayOutputStream
+        val byteArrayOutputStream = ByteArrayOutputStream()
+        croppedBitmap.compress(Bitmap.CompressFormat.JPEG, quality, byteArrayOutputStream)
+
+        // return ByteArray of compressed JPEG image (representing .jpg)
+        byteArrayOutputStream.toByteArray()
+    } catch (e: Exception) {
+        e.printStackTrace()
+        null
     }
+}
 }
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -4,8 +4,8 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.net.Uri
-import android.widget.Toast
 import android.util.Log
+import android.widget.Toast
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import com.google.firebase.Firebase
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -159,21 +159,22 @@ open class ProfilesViewModel(
         userId = uid, onSuccess = {}, onFailure = { e -> handleError(e) })
   }
 
-/**
- * Uploads the current user's profile picture to the remote storage system.
- *
- * @param imageData The byte array representing the image data to be uploaded.
- * @param onSuccess A callback function that is invoked with the URL of the uploaded image upon successful upload.
- * @throws IllegalStateException if the user is not logged in.
- */
-fun uploadCurrentUserProfilePicture(imageData: ByteArray, onSuccess: (url: String?) -> Unit) {
-  val userId = selectedProfile.value?.uid ?: throw IllegalStateException("User not logged in")
-  ppRepository.uploadProfilePicture(
-      userId = userId,
-      imageData = imageData,
-      onSuccess = { url -> onSuccess(url) },
-      onFailure = { e -> handleError(e) })
-}
+  /**
+   * Uploads the current user's profile picture to the remote storage system.
+   *
+   * @param imageData The byte array representing the image data to be uploaded.
+   * @param onSuccess A callback function that is invoked with the URL of the uploaded image upon
+   *   successful upload.
+   * @throws IllegalStateException if the user is not logged in.
+   */
+  fun uploadCurrentUserProfilePicture(imageData: ByteArray, onSuccess: (url: String?) -> Unit) {
+    val userId = selectedProfile.value?.uid ?: throw IllegalStateException("User not logged in")
+    ppRepository.uploadProfilePicture(
+        userId = userId,
+        imageData = imageData,
+        onSuccess = { url -> onSuccess(url) },
+        onFailure = { e -> handleError(e) })
+  }
 
   /**
    * Deletes the current user's profile picture from the remote storage system.
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -223,8 +223,10 @@ open class ProfilesViewModel(
   private fun imageUriToJpgByteArray(
       context: Context,
       imageUri: Uri,
-      quality: Int = 100
+      quality: Int = 100,
+      maxResolution: Int = 600
   ): ByteArray? {
+
     return try {
       // Open an InputStream from the URI
       val inputStream: InputStream? = context.contentResolver.openInputStream(imageUri)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfilePictureSelector.kt
```diff
@@ -0,0 +1,98 @@
...
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import coil.compose.AsyncImage
+import com.github.se.icebreakrr.R
+import com.github.se.icebreakrr.ui.theme.IceBreakrrBlue
+
+@Preview(showBackground = true)
```
louiscrc, line 30: nitpick : don't forget to remove previews and dev-related code for the PR

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -186,6 +212,47 @@ open class ProfilesViewModel(
         onFailure = { e -> handleError(e) })
   }
 
+  /**
```
louiscrc, line 215: Just an idea of implementation: when you select a picture on your phone, you might want to be able to crop the part of the picture you want to rather than the square center of the picture. Libraries like [Android Image Cropper](https://github.com/CanHub/Android-Image-Cropper) seem to allow this. You could try to implement this in M3.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
```diff
@@ -126,14 +139,16 @@ fun ProfileEditingScreen(
...
-                            .background(MaterialTheme.colorScheme.primary)
-                            .testTag("profilePicture"))
+                // A composable that allows the user to preview and edit a profile picture
+                ProfilePictureSelector(
+                    url = user.profilePictureUrl,
+                    size = profilePictureSize,
+                    onSelectionSuccess = { uri ->
+                      profilesViewModel.processAndUploadImage(context, uri)
+                    },
+                    onSelectionFailure = {
```
louiscrc, line 149: This appears even when I cancel the picture selection (which is technically not a failure though)

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -186,6 +212,47 @@ open class ProfilesViewModel(
...
+   * Converts an image URI to a JPEG byte array, cropping the image to a square at the center.
+   *
+   * @param context The context used to access the content resolver.
+   * @param imageUri The URI of the image to be converted.
+   * @param quality The quality of the JPEG compression (0-100). Default is 100.
+   * @return A byte array representing the JPEG image, or null if an error occurs.
+   */
+  private fun imageUriToJpgByteArray(
```
louiscrc, line 223: One security issue seems to appear : if the user uploads an image that is very large and heavy, it would still be uploaded to the database.
I think the image should be reduced to a fixed size (like 600x600) to prevent this.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
louiscrc, line 1: Based on your modifications, the `isLoading` variable on line 77 should be different, and it should only consider loading when first loading the page. Updating the profile URI doesn't need to display the `isLoading` version of the page.

louiscrc (changes requested):
Nice PR! It is essential for a user to be able to choose their profile picture, and you implemented this functionality very well.

Apart from the comments that are generally more suggestions, I think the security issue regarding image size should be fixed before this PR is merged. 

Fixing the loading state when changing the profile picture would also be nice.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfilePictureSelector.kt
Kapestian: Oh, I did it 30s ago 😂

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
Kapestian: Yes, This was planned, but I didn't know how. I'll take a look at your recommendation at that moment. Thanks!

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
Kapestian: As I said in the Description of the PR, this was made on purpose to avoid problems. I may have found a way to make discarding work anyway for the profile picture, so I'll ask the rest of the team and change if the majority wants it.

Kapestian: The more I think of it, the more I think it is right to discard the picture with the rest, even if it requires more work. I am currently working on a fix for that

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
Kapestian: You are absolutely right. I don't think this is a security problem but it definitely does not help with the slow upload speed. This is now fixed

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
Kapestian: the problem is that i am calling `updateUser` in the `profilesViewModel` which briefly turns `isLoading` to `true`. I am working on a fix to make the profile picture discardable, this issue will be disappear at the same time

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
Kapestian: Ok, this is now fixed and clean, but now we have to solve glitches appearing in the `ProfileView` screen. I have some solutions to discuss during our next meeting (Monday)

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
Kapestian: update: the problem is not visible in this screen anymore, but it appears on the `ProfileView` screen now as said in previous discussion.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
```diff
@@ -126,14 +141,18 @@ fun ProfileEditingScreen(
...
+                    url = user.profilePictureUrl,
+                    localBitmap = tempBitmap,
+                    size = profilePictureSize,
+                    onSelectionSuccess = { uri ->
+                      profilesViewModel.generateTempProfilePictureBitmap(context, uri)
+                      isModified = true
+                    },
+                    onSelectionFailure = {
+                      Toast.makeText(context, "Failed to select image", Toast.LENGTH_SHORT).show()
+                    })
```
Kapestian, line 155: help: @louiscrc this is the last lines I have to test to reach 80% coverage but after 1hours on it I still can't find how to test them in a useful way, could you help me?

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -31,6 +38,9 @@ open class ProfilesViewModel(
   private val _error = MutableStateFlow<Exception?>(null)
   val error: StateFlow<Exception?> = _error
 
+  private val _tempProfilePictureBitmap = MutableStateFlow<Bitmap?>(null)
```
louiscrc, line 41: Your implementation of a temporary local profile picture is smart! This indeed reduces the loading time.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -197,10 +244,74 @@ open class ProfilesViewModel(
...
+   *
+   * @param context The context used to access the content resolver.
+   * @param imageUri The URI of the image to be converted.
+   * @param maxResolution The maximum resolution for the output Bitmap. Default is 600.
+   * @return A Bitmap representing the processed image, or null if an error occurs.
+   */
+  private fun imageUriToBitmap(context: Context, imageUri: Uri, maxResolution: Int = 600): Bitmap? {
+    return try {
+      // Open an InputStream from the URI
+      val inputStream: InputStream? = context.contentResolver.openInputStream(imageUri)
```
louiscrc, line 263: Considering further implementation of a cropping library, this will have to be modified : you don't want to strictly crop a 600x600 frame out of the original image, but rather let the user choose the part he wants, and then degrade the quality of its selection to obtain a 600x600 version of what he selected.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
louiscrc: Gepeto suggets to use a `remember` and `derivedStateOf` to ensure `isLoading` remains `false` after the initial change. This approach could work, as we only need to display the loading screen when the entire profile has not been loaded.

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
Kapestian: NVM, the actual line coverage is 80%. I was reading the weird coverage displayed on GitHub which is not really the line coverage

louiscrc (approved):
Nice work Valentin! You implemented my suggestions well, and prepared the field well for the image cropping functionality. 
I can't wait to replace my profile picture of a cat by another cat 🐱

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
Kapestian: I would be interesting to take a look later. But as said before, this is not linked to the profile picture features anymore

PR #69, opened 2024-11-10
Feature/authentication : the logout

JanStaszewiczEPFL:
### The logout

This PR introduces the logout functionality into the app, along with several related updates and improvements to streamline the implementation and testing process.

**Details**

**Logout Functionality:**

Created a `User.kt` file inside the authentication folder to keep the authentication-related logic clean and organized.
Implemented a logout function in `User.kt` that:

1. Logs the user out.
2. Resets Google Sign-In options.
3. Navigates the user back to the sign-in screen.

Connected this function to the logout button in the `Settings.kt` file.

**Navigation Tests:**

Added navigation tests for the `Settings.kt` file to validate the logout functionality.
However, due to a conflict with a commented version of `Settings.kt` (pushed and merged by @MaxCastelCast), I modified the tests to work with my implementation but left them commented out as agreed upon with the team even though I don't feel like it is the right way to proceed. Then I merged with the new `NavigationTest.kt` file but I couldn't implement my tests are the way of testing the navigation completely changed. 

**Test Cleanup:**

Removed unnecessary tests in `FirebaseAuthTest.kt` to reduce redundancy and improve test suite maintainability.

As for now, we find unnecessary (also discussed and agreed with the coaches) to test the google functions as there is no straight forward way to do so.

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/authentication/User.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/authentication/User.kt
@@ -24,30 +24,30 @@ import com.google.firebase.auth.FirebaseAuth
  */
 fun logout(context: Context, navigationActions: NavigationActions) {
 
-    //Initialize FirebaseAuth
-    val auth = FirebaseAuth.getInstance()
+  // Initialize FirebaseAuth
+  val auth = FirebaseAuth.getInstance()
 
-    //Configure GoogleSignIn
-    val gso =
-        GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
-            .requestIdToken(
-                context.getString(R.string.default_web_client_id)) // Use Web client ID as a string
-            .requestEmail()
-            .build()
+  // Configure GoogleSignIn
+  val gso =
+      GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
+          .requestIdToken(
+              context.getString(R.string.default_web_client_id)) // Use Web client ID as a string
+          .requestEmail()
+          .build()
 
-    val googleSignInClient: GoogleSignInClient = GoogleSignIn.getClient(context, gso)
+  val googleSignInClient: GoogleSignInClient = GoogleSignIn.getClient(context, gso)
 
-    //Sign out from Firebase
-    auth.signOut()
+  // Sign out from Firebase
+  auth.signOut()
 
-    //Sign out from Google and reset the google account choice
-    googleSignInClient.signOut().addOnCompleteListener { signOutTask ->
-        if (signOutTask.isSuccessful) {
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/icebreakrr/authentication/User.kt
```diff
@@ -0,0 +1,53 @@
...
+
+  val googleSignInClient: GoogleSignInClient = GoogleSignIn.getClient(context, gso)
+
+  // Sign out from Firebase
+  auth.signOut()
+
+  // Sign out from Google and reset the google account choice
+  googleSignInClient.signOut().addOnCompleteListener { signOutTask ->
+    if (signOutTask.isSuccessful) {
+      googleSignInClient.revokeAccess().addOnCompleteListener { revokeTask ->
```
MaxCastelCast, line 46: Here the SonarCloud says that the unused function parameter is a major maintanability issue, so it should be removed and replaced with _ , we ll then need to proceed everywhere else in our code where we have unused lambda variables and remove them

app/src/main/java/com/github/se/icebreakrr/ui/sections/Settings.kt
```diff
@@ -89,7 +93,12 @@ fun SettingsScreen(profilesViewModel: ProfilesViewModel, navigationActions: Navi
...
               onClick = {
-                Toast.makeText(context, "Log Out (Not yet implemented)", Toast.LENGTH_SHORT).show()
+                if (isTesting) {
+                  navigationActions.navigateTo(Screen.AUTH)
+                } else {
+                  auth = FirebaseAuth.getInstance()
+                  auth.currentUser?.let { logout(context, navigationActions) }
+                }
               },
               colors = ButtonDefaults.buttonColors(containerColor = Color(0xFFCE0E00)),
```
MaxCastelCast, line 103: Hardcoded color, should be classified in Color.kt and use the constant from there for more maintanability

MaxCastelCast (changes requested):
The code is globally correct, but i have requested some changes that were highlighted by the SonarCloud and some constants issues, alro i wanted to ask if you could let me push py code before yours since i have re-written the tests for NavAction and EndToEnd1 and they work localy, maybe it would be better for you to push with those test already in your code, sorry for the inconvenience

app/src/main/java/com/github/se/icebreakrr/ui/sections/Settings.kt
JanStaszewiczEPFL: I didn't work on Settings.kt but I can fix that sure

app/src/main/java/com/github/se/icebreakrr/authentication/User.kt
JanStaszewiczEPFL: You are right, I will change that!

MaxCastelCast (approved):
Looks good to me !

PR #117, opened 2024-11-12
Feature/datastore

BotondAKovacs:
# DataStore Implementation

## Overview
This PR implements persistent data storage using DataStore to enhance offline functionality.

## Key Changes
1. **DataStore Implementation**
   - Created DataStore class for handling auth token and discoverability preferences
   - Added methods for saving, retrieving, and clearing user data
   - Implemented unit tests for DataStore operations
  

3. **Network Utilities**
   -  Modified it so that isNetworkAvailable doesn't require context anymore

4. **Integration**
   - Added DataStore initialization in app startup

5. **New Features**
  - Added a launch effect to the AroundYou page so that when the user logs in offline it immediately displays "no internet" instead of having to wait for the timeout
   - Force firebase to retrieve from cache if the user is offline so that the user can immediately see its profile
   - If the user has already signed in once, the auth token is stored and the user can next time login offline

## Technical Details
- Uses Preferences DataStore for persistent storage

## Testing
- Unit tests for DataStore operations
- Network utility tests with mock ConnectivityManager
- Integration tests for offline functionality

## Dependencies
- Added DataStore dependencies to Gradle

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/data/AppDataStore.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/data/AppDataStore.kt
@@ -1,0 +1,92 @@
+package com.github.se.icebreakrr.data
+
+import android.content.Context
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.booleanPreferencesKey
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.stringPreferencesKey
+import androidx.datastore.preferences.preferencesDataStore
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+/**
+ * Manages persistent app preferences using DataStore. This class handles storage and retrieval of
+ * user preferences such as:
+ * - Authentication token
+ * - Discoverability status
+ *
+ * All operations are suspending functions or return Flows to ensure safe background execution.
+ */
+class AppDataStore(private val dataStore: DataStore<Preferences>) {
+  /**
+   * Secondary constructor that creates a DataStore instance from a Context.
+   *
+   * @param context The application context used to create the DataStore
+   */
+  constructor(context: Context) : this(context.dataStore)
+
+  companion object {
+    /**
+     * DataStore instance created using the preferencesDataStore delegate. Ensures only one instance
+     * of DataStore is created for the application.
+     */
+    private val Context.dataStore: DataStore<Preferences> by
+        preferencesDataStore(name = "app_preferences")
+
+    // Preference Keys
+    /** Key for storing the authentication token Type: String */
+    private val AUTH_TOKEN_KEY = stringPreferencesKey("auth_token")
+
+    /** Key for storing the user's discoverability preference Type: Boolean */
+    private val DISCOVERABLE_KEY = booleanPreferencesKey("is_discoverable")
+  }
+
+  /**
+   * Provides access to the stored authentication token as a Flow.
+   *
+   * @return Flow<String?> that emits the current auth token or null if not set
+   */
+  val authToken: Flow<String?> = dataStore.data.map { preferences -> preferences[AUTH_TOKEN_KEY] }
+
+  /**
+   * Saves an authentication token to persistent storage.
+   *
+   * @param token The authentication token to store
+   */
+  suspend fun saveAuthToken(token: String) {
+    dataStore.edit { preferences -> preferences[AUTH_TOKEN_KEY] = token }
+  }
+
+  /**
+   * Removes the stored authentication token. Used during logout or when the token needs to be
+   * invalidated.
+   */
+  suspend fun clearAuthToken() {
+    dataStore.edit { preferences -> preferences.remove(AUTH_TOKEN_KEY) }
+  }
+
+  /**
+   * Clears all stored preferences. Use with caution as this will reset all preferences to their
+   * default values.
+   */
+  suspend fun clearAll() {
+    dataStore.edit { preferences -> preferences.clear() }
+  }
+
+  /**
+   * Check if we have a valid auth token stored
+   *
+   * @return Flow<Boolean> indicating if we have a token
+   */
+  val hasAuthToken: Flow<Boolean> = authToken.map { token -> !token.isNullOrEmpty() }
+
+  /** Save discoverability preference */
+  suspend fun saveDiscoverableStatus(isDiscoverable: Boolean) {
+    dataStore.edit { preferences -> preferences[DISCOVERABLE_KEY] = isDiscoverable }
+  }
+
+  /** Get discoverability status as a Flow Default value is true to maintain existing behavior */
+  val isDiscoverable: Flow<Boolean> =
+      dataStore.data.map { preferences -> preferences[DISCOVERABLE_KEY] ?: true }
+}
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
@@ -3,12 +3,14 @@ package com.github.se.icebreakrr.model.profile
 import android.os.Handler
 import android.os.Looper
 import android.util.Log
+import com.github.se.icebreakrr.utils.NetworkUtils
 import com.google.android.gms.tasks.Task
 import com.google.firebase.Firebase
 import com.google.firebase.auth.auth
 import com.google.firebase.firestore.DocumentSnapshot
 import com.google.firebase.firestore.FirebaseFirestore
 import com.google.firebase.firestore.GeoPoint
+import com.google.firebase.firestore.Source
 import kotlinx.coroutines.flow.MutableStateFlow
 
 class ProfilesRepositoryFirestore(private val db: FirebaseFirestore) : ProfilesRepository {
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/github/se/icebreakrr/data/AppDataStore.kt
```diff
@@ -0,0 +1,94 @@
...
+   */
+  val hasAuthToken: Flow<Boolean> = authToken.map { token -> !token.isNullOrEmpty() }
+
+  /** Save discoverability preference */
+  suspend fun saveDiscoverableStatus(isDiscoverable: Boolean) {
+    dataStore.edit { preferences -> preferences[DISCOVERABLE_KEY] = isDiscoverable }
+  }
+
+  /** Get discoverability status as a Flow Default value is true to maintain existing behavior */
+  val isDiscoverable: Flow<Boolean> =
```
arthur-mrgt, line 92: Nitpick: This should be placed before the functions (see the wiki 3. Style Convention/Declaration Order).

app/src/main/java/com/github/se/icebreakrr/data/AppDataStore.kt
```diff
@@ -0,0 +1,94 @@
...
+      dataStore.data.map { preferences -> preferences[DISCOVERABLE_KEY] ?: true }
```
arthur-mrgt, line 93: Nitpick: Defaulting to true is a reasonable approach. However, this hardcoded default could be extracted to a constant if it is referenced elsewhere for maintainability.

app/src/main/java/com/github/se/icebreakrr/data/AppDataStore.kt
```diff
@@ -0,0 +1,94 @@
...
+   */
+  val authToken: Flow<String?> = dataStore.data.map { preferences -> preferences[AUTH_TOKEN_KEY] }
+
+  /**
+   * Saves an authentication token to persistent storage.
+   *
+   * @param token The authentication token to store
+   */
+  suspend fun saveAuthToken(token: String) {
+    dataStore.edit { preferences -> preferences[AUTH_TOKEN_KEY] = token }
```
arthur-mrgt, line 60: Important: It would benefit from handling empty strings or invalid tokens by adding a check before saving to avoid storing invalid data.

app/src/main/java/com/github/se/icebreakrr/data/AppDataStore.kt
```diff
@@ -0,0 +1,94 @@
...
+     */
+    private val Context.dataStore: DataStore<Preferences> by
+        preferencesDataStore(name = "app_preferences")
+
+    // Preference Keys
+    /** Key for storing the authentication token Type: String */
+    private val AUTH_TOKEN_KEY = stringPreferencesKey("auth_token")
+
+    /** Key for storing the user's discoverability preference Type: Boolean */
+    private val DISCOVERABLE_KEY = booleanPreferencesKey("is_discoverable")
```
arthur-mrgt, line 44: Nitpick: AUTH_TOKEN_KEY  & DISCOVERABLE_KEY  instances are objects created at runtime, so they do not qualify as a const val. Therefore using camelCase (authTokenKey or discoverableKey) instead of uppercase (AUTH_TOKEN_KEY) aligns more with our convention.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
```diff
@@ -202,8 +204,14 @@ class ProfilesRepositoryFirestore(private val db: FirebaseFirestore) : ProfilesR
       onSuccess: (Profile?) -> Unit,
       onFailure: (Exception) -> Unit
   ) {
+    val source =
```
arthur-mrgt, line 207: Great solution!

app/src/main/java/com/github/se/icebreakrr/ui/authentication/SignIn.kt
```diff
@@ -193,8 +205,19 @@ fun SignInScreen(
                 if (isTesting) {
                   navigationActions.navigateTo(TopLevelDestinations.AROUND_YOU)
                 } else {
-                  if (!NetworkUtils.isNetworkAvailable(context)) {
-                    NetworkUtils.showNoInternetToast(context)
+                  if (!NetworkUtils.isNetworkAvailable()) {
+                    if (hasAuthToken.value) {
+                      // If offline but has token, allow access
+                      navigationActions.navigateTo(TopLevelDestinations.AROUND_YOU)
+                      profilesViewModel.getFilteredProfilesInRadius(
```
arthur-mrgt, line 212: Question: Why calling this ? If it is in offline mode, nobody should be displayed in the aroundYou page.

app/src/main/java/com/github/se/icebreakrr/ui/sections/AroundYou.kt
```diff
@@ -63,6 +65,28 @@ fun AroundYouScreen(
   val context = LocalContext.current
   val isConnected = profilesViewModel.isConnected.collectAsState()
 
+  // Define constants for magic numbers
+  val defaultGeoPoint = GeoPoint(0.0, 0.0)
+  val searchRadius = 300.0
+  val verticalPadding = 16.dp
+  val horizontalPadding = 8.dp
+  val messageTextSize = 20.sp
+  val messageTextColor = Color(0xFF575757)
```
arthur-mrgt, line 74: Nitpick: If they are constants, use const val and write the names in uppercase.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Settings.kt
```diff
@@ -44,14 +45,29 @@ import com.github.se.icebreakrr.ui.sections.shared.TopBar
...
+
+  // Collect the discoverability state from DataStore
+  val isDiscoverable by appDataStore.isDiscoverable.collectAsState(initial = true)
+
   lateinit var auth: FirebaseAuth
 
+  // Constants for padding and spacing
+  val screenPadding = 16.dp
+  val verticalSpacing = 8.dp
+  val buttonVerticalSpacing = 16.dp
```
arthur-mrgt, line 69: Nitpick: If they are constants, use const val and write the names in uppercase.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Settings.kt
```diff
@@ -114,22 +140,31 @@ fun SettingsScreen(profilesViewModel: ProfilesViewModel, navigationActions: Navi
...
+    label: String,
+    isChecked: Boolean,
+    onCheckedChange: (Boolean) -> Unit,
+    modifier: Modifier = Modifier
+) {
+  // Constants for card styling
+  val cardShape = RoundedCornerShape(16.dp)
+  val cardElevation = 4.dp
+  val cardHeight = 55.dp
+  val cardPadding = 16.dp
```
arthur-mrgt, line 153: Nitpick: If they are constants, use const val and write the names in uppercase.

app/src/main/java/com/github/se/icebreakrr/ui/authentication/SignIn.kt
BotondAKovacs: If I remember correctly, I did this because if I don't, when you signin without connectivity, the function that tries to reconnect every 5 seconds isn't immediately launched. But thanks to your comment I realized that my implementation is faulty because I set isConnected to false on launch in arroundYou if the user has no connectivity which stops the reconnection function from launching anyways.

arthur-mrgt (changes requested):
Overall, this is a very good PR. The requested features are well implemented, and the code is clear and well-documented. I've noted a few minor issues that need to be addressed.

If I had to make one suggestion, it would be that the AppDataStore class could benefit from being more modular. For example, we could have functions like:

// Retrieve a Flow of a stored preference by key, with a default value.
fun <T> getPreference(key: Preferences.Key<T>, defaultValue: T): Flow<T> =
    dataStore.data.map { preferences -> preferences[key] ?: defaultValue }

// Add or update a preference with a given key and value.
suspend fun <T> putPreference(key: Preferences.Key<T>, value: T) {
    dataStore.edit { preferences -> preferences[key] = value }
}

// Delete a specific preference by its key.
suspend fun <T> removePreference(key: Preferences.Key<T>) {
    dataStore.edit { preferences -> preferences.remove(key) }
}

// Clear all stored preferences.
suspend fun clearAllPreferences() {
    dataStore.edit { preferences -> preferences.clear() }
}

This would make the class more modular, as currently, adding new preferences requires adding specific functions. For common use cases, we could simply add code like:

// Define some specific preference keys for common use cases.
private val authTokenKey = stringPreferencesKey("auth_token")
private val discoverableKey = booleanPreferencesKey("is_discoverable")

// Commonly used preference Flows
val authToken: Flow<String?> = getPreference(authTokenKey, null)
val isDiscoverable: Flow<Boolean> = getPreference(discoverableKey, true)

// Commonly used preference functions for convenience.
suspend fun saveAuthToken(token: String) = putPreference(authTokenKey, token)
suspend fun clearAuthToken() = removePreference(authTokenKey)
suspend fun saveDiscoverableStatus(isDiscoverable: Boolean) = putPreference(discoverableKey, isDiscoverable)

// Check if auth token exists
val hasAuthToken: Flow<Boolean> = authToken.map { !it.isNullOrEmpty() }

In addition to generic functions, this approach would make future implementations (such as saving filters) simpler to add.

If, however, this class is only used for two or three specific cases, it may not be necessary to make it modular.

Otherwise, the tests are well done, though they could benefit from broader coverage.

app/src/main/java/com/github/se/icebreakrr/ui/sections/AroundYou.kt
BotondAKovacs: I believe in the meantime Maximo did a PR fixing the magic numbers in these 2 screens, I'll merge and accept his when he merges with the main. But you are right.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Settings.kt
BotondAKovacs: I believe in the meantime Maximo did a PR fixing the magic numbers in these 2 screens, I'll merge and accept his when he merges with the main. But you are right.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
BotondAKovacs: Thanks!

BotondAKovacs:
I started moving magic numbers into correct const vals but realised that that was someone else's job but I have to merge now, I'll let them fix it. Also I changed a UI constant (size of tag box in profile view) to make it look nicer because I noticed it looked weird

PR #118, opened 2024-11-13
feature : block-backend

louiscrc:
# New Feature: Block Users

##### (I already blocked @BotondAKovacs after his last review on my PR)

---

## **How does it work?**

The `Profile` data class has been updated to include a `hasBlocked` field, which is a `List<String>` containing the user IDs of any users you have blocked.

### **Implementation Details:**
- **ProfilesViewModel Update:** 
  - The `ProfilesViewModel` has been enhanced to simultaneously retrieve the displayed user and the current phone's user.
  - To refresh the user's own profile, call `profilesViewModel.getSelfProfile()`. This will update `profilesViewModel.selfProfile` with the `Profile` of the user on the device (retrieved from `Firebase.auth.uid`).

- **Blocking Method:**
  - The blocking functionality is implemented in `ProfilesViewModel`. It works by updating the `hasBlocked` list of the phone's user `Profile` to include the `selectedProfile`. 

### ⚠️ **Caution:**
- **One-Way Blocking Only:** Currently, blocking a user is only one-way. There's no screen yet to view or unblock users you've blocked. You have to remove them manually in Firebase.

---

## **Next Steps:**
- Adapt `ProfileView` and `ProfileEdit` to directly use `selfProfile` instead of `selectedProfile`.
- Update the tests to reflect these changes, and potentially add more tests to improve coverage. (Our current line coverage is 86%, aiming for 💯)
- Implement an unblock screen (planned for Milestone 3).
- Do the implementation of the report-user (which is slightly more complicated, but this PR prepares the field well)

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/Profile.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/Profile.kt
@@ -24,7 +24,8 @@ data class Profile(
     val description: String,
     val tags: List<String> = listOf(),
     val profilePictureUrl: String? = null,
-    val fcmToken: String? = null
+    val fcmToken: String? = null,
+    var hasBlocked: List<String> = listOf(),
 ) {
   /**
    * Calculates the user's age based on their birth date.
--- a/app/src/androidTest/java/com/github/se/icebreakrr/ui/profile/OtherProfileViewTest.kt
+++ b/app/src/androidTest/java/com/github/se/icebreakrr/ui/profile/OtherProfileViewTest.kt
@@ -171,6 +171,7 @@ class OtherProfileViewTest {
   fun testBlockUserFlowWithConfirm() {
     fakeProfilesViewModel.setLoading(false)
     fakeProfilesViewModel.setSelectedProfile(Profile.getMockedProfiles()[0])
+    fakeProfilesViewModel.setSelfProfile(Profile.getMockedProfiles()[1])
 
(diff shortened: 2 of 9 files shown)


louiscrc:
You can already review my current implementation, this is the chore of the PR.
Any further commits is about adapting the rest of the code to my new implementation, and adapting / creating tests.

app/src/main/java/com/github/se/icebreakrr/model/profile/Profile.kt
```diff
@@ -24,7 +24,8 @@ data class Profile(
     val description: String,
     val tags: List<String> = listOf(),
     val profilePictureUrl: String? = null,
-    val fcmToken: String? = null
+    val fcmToken: String? = null,
+    var hasBlocked: List<String> = listOf(),
```
BotondAKovacs, line 28: You could use a Set instead of a list, since each user can only be once in it.

BotondAKovacs (changes requested):
Nice work! But I have a few issues:
Please pay attention to respecting the branch naming conventions (using feature/...)
I also found a bug, if you block someone, the first time you quit the app and come back you can still see the blocked users until a refresh is performed. This should be fixed.
 I understand that you didn't have time to also implement the list view of users blocked by the user, but at least for testing purposes, you could add a "deblock all" button in the settings.

louiscrc:
This PR was closed as I renamed the branch correctly.

PR #127, opened 2024-11-14
Fix/navigation fix

Kapestian:
# Navigation fix
## what's new
This pull request is a simple fix to the two following bugs:
* The phone's "back button" was bypassing the confirmation popup
* navigating from a top level screen to another was possible by tapping on the phone's "back button"
I added tests for new code (and also completed the ProfileEdit test suit)

## Implementation
I created the utility function `handleSafeBackNavigation` in [section.shared.NavigationUtils.kt](https://github.com/Swent-team-6/icebreakrr/blob/2769d0391a968b82c62674b24507c455fa564e21/app/src/main/java/com/github/se/icebreakrr/ui/sections/shared/NavigationUtils.kt) file to generalize the behavior of going back when a confirmation may appear. This function is then used both in `ProfileEdit.kt` and `Filter.kt` when clicking on our app's back arrow and in a so called `BackHandler` that adds the desired behavior to the system "back button"

## To go further 
It may be a good idea to implement the "press back again to quit the app" Toast to avoid frustration, because right now going back from a top level almost feels like getting catapulted out of the app (it is the desired navigation behavior, but it could be better).

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/ui/navigation/NavigationAction.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/navigation/NavigationAction.kt
@@ -1,6 +1,5 @@
 package com.github.se.icebreakrr.ui.navigation
 
-import androidx.navigation.NavGraph.Companion.findStartDestination
 import androidx.navigation.NavHostController
 
 open class NavigationActions(
--- a/app/src/main/java/com/github/se/icebreakrr/ui/navigation/NavigationAction.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/navigation/NavigationAction.kt
@@ -15,7 +15,7 @@ open class NavigationActions(
    */
   open fun navigateTo(destination: TopLevelDestination) {
     navController.navigate(destination.route) {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/se/icebreakrr/ui/navigation/NavigationAction.kt
```diff
@@ -15,7 +14,7 @@ open class NavigationActions(
    */
   open fun navigateTo(destination: TopLevelDestination) {
     navController.navigate(destination.route) {
```
Kapestian, line 16: Note: this is the way the bootcamp was coded

louiscrc (approved):
Unsatisfying PR when it's too good and I can't make remarks :/

The only thing that comes to my mind is that you could try to make a custom `BackHandler` component it's always the same format of
 ```
 BackHandler {
    handleSafeBackNavigation(
        isModified = isModified,
        setShowDialog = { showDialog = it },
        tagsViewModel = tagsViewModel,
        navigationActions = navigationActions)
  }
  ```

PR #129, opened 2024-11-14
Feature/profiles : Profile Creation Screen On First Sign In

JanStaszewiczEPFL:
# Profile Creation Integration

## Changes Made

**Summary :** adding profile creation on first sign in and implementing it into app flow.

### Integrated Profile creation screen

- Added a `ProfileCreationScreen` composable that allows the user to customize profile on first sign in. See file `ProfileCreation.kt`.
- The screen was implemented into the app flow in the `MainActivity.kt` and in the navigation system.
- The SignIn.kt file was changed to integrate navigation to the `ProfileCreationScreen` if there was no `IceBreakrr` account for the user. 
- The Firebase FCM token generation was moved from `SignIn.kt` to the `ProfileCreationScreen` for this implementation. 

### Date Handling Standardization

- Created a `DateUtils` file in the `utils` folder to provide some basic date handling functions.

### Test Coverage

- Implemented tests for `ProfileCreation.kt`.
- Implemented tests for `DateUtils.kt`.

image

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
@@ -79,11 +79,9 @@ fun ProfileCreationScreen(
   val catchphrase = remember { mutableStateOf("") }
   val description = remember { mutableStateOf("") }
 
-  val tempBitmap = profilesViewModel.tempProfilePictureBitmap.collectAsState().value
-
   val selectedTags = tagsViewModel.filteringTags.collectAsState().value
   val tagsSuggestions = tagsViewModel.tagsSuggestions.collectAsState()
-  val stringQuery = tagsViewModel.query.collectAsState()
+  val stringQuery = remember { mutableStateOf("") }
 
   val expanded = remember { mutableStateOf(false) }
 
@@ -265,7 +263,7 @@ fun ProfileCreationScreen(
               selectedTag = selectedTags.map { tag -> Pair(tag, tagsViewModel.tagToColor(tag)) },
               outputTag =
                   tagsSuggestions.value.map { tag -> Pair(tag, tagsViewModel.tagToColor(tag)) },
-              stringQuery = stringQuery.value,
+              stringQuery = stringQuery,
               expanded = expanded,
               onTagClick = { tag ->
                 tagsViewModel.removeFilter(tag)
--- a/app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
@@ -1,0 +1,314 @@
+import android.app.DatePickerDialog
+import android.widget.Toast
+import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Check
+import androidx.compose.material.icons.filled.DateRange
+import androidx.compose.material3.*
+import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButton
+import androidx.compose.material3.OutlinedTextField
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.github.se.icebreakrr.model.profile.Gender
+import com.github.se.icebreakrr.model.profile.Profile
+import com.github.se.icebreakrr.model.profile.ProfilesViewModel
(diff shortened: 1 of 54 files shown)


app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
```diff
@@ -0,0 +1,314 @@
...
+        description.value.isNotBlank() &&
+        selectedGender != null
+  }
+
+  // Create DatePicker
+  val datePicker =
+      DatePickerDialog(
+          context,
+          { _, year, month, dayOfMonth ->
+            val newCalendar =
```
louiscrc, line 104: You use twice the same instance of calendar
Copy it into one variable and reuse the variable

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
```diff
@@ -0,0 +1,314 @@
...
+  val user = profilesViewModel.selectedProfile.collectAsState().value
+
+  val selectedTags = tagsViewModel.filteringTags.collectAsState().value
+  val tagsSuggestions = tagsViewModel.tagsSuggestions.collectAsState()
+  val stringQuery = tagsViewModel.query.collectAsState()
+
+  val expanded = remember { mutableStateOf(false) }
+
+  var showDialog by remember { mutableStateOf(false) }
+  var isModified by remember { mutableStateOf(false) }
```
louiscrc, line 70: You sometimes set this var to true, but never actually use it. Is this intentional ?

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
```diff
@@ -0,0 +1,314 @@
...
+  val defaultDate = calendar.time
+
+  val birthdate = remember { mutableStateOf(DateUtils.formatDate(defaultDate)) }
+  val selectedDate = remember { mutableStateOf(defaultDate) }
+  var isDateValid = remember { mutableStateOf(true) }
+
+  var selectedGender by remember { mutableStateOf<Gender?>(null) }
+
+  // Add form validation
+  fun isFormValid(): Boolean {
```
louiscrc, line 90: As suggested by Botond, we might want to force the user to select at least x tags (you can create a variable MIN_REQUIRED_TAGS that we can modify then), and so you should check that the user has selected at least this minimum.

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
louiscrc, line 1: Global nitpick: there are a lot of magic numbers on this file. Use fixed vars instead.

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
```diff
@@ -0,0 +1,314 @@
...
+                      Profile(
+                          uid = user.uid,
+                          name = fullName.value,
+                          gender = selectedGender!!, // Add this line
+                          birthDate =
+                              Timestamp(
+                                  Calendar.getInstance()
+                                      .apply {
+                                        time = selectedDate.value!!
+                                        set(Calendar.HOUR_OF_DAY, 0)
```
louiscrc, line 274: I think you have already set hour/minute/seconds/milliseconds on the pickedDate variable, so no need to set it again.

louiscrc (changes requested):
Globally a very nice PR, essential in the user path.
I'll request changes just for the magic numbers, and so that you can check my remarks on certain portions of code.

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
```diff
@@ -0,0 +1,314 @@
...
+                                        set(Calendar.MILLISECOND, 0)
+                                      }
+                                      .time),
+                          catchPhrase = catchphrase.value,
+                          description = description.value,
+                          tags = selectedTags,
+                          fcmToken = fcmToken)
+                  profilesViewModel.addNewProfile(newProfile)
+                  // Upload profile picture if selected
+                  profilesViewModel.validateAndUploadProfilePicture(context)
```
Kapestian, line 286: Bug: to use the validateAndUploadProfilePicture you must set `selectedProfile` of `profilesViewModel` to the profile you just created, see why below:

image

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
JanStaszewiczEPFL: Yes I'll change that!

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
JanStaszewiczEPFL: No I will remove showDialog

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
JanStaszewiczEPFL: As for now this isn't necessary and the team didn't agree on this dynamic

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
JanStaszewiczEPFL: Yes you are right I will fix that

app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
JanStaszewiczEPFL: yep!

louiscrc (approved):
Nice PR ! LGTM

PR #136, opened 2024-11-14
Hotfix/issues fixing

MaxCastelCast:
Fix some issues : 
-TagSelector drop down menu items where too big
-sometimes the profiles where deleted when the app is uninstalled/reinstalled (Valentin helped me a lot)
-removed constants in the Sign-in page
-confirm that the documentation is complete
-Tags in the view model are now shown
-Selecting tags in the profile now is independent from the one of filter (Sam debugged this part, I just wrote it)

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
@@ -270,6 +270,7 @@ fun ProfileCreationScreen(
                 isModified = true
               },
               onStringChanged = {
+                stringQuery.value = it
                 tagsViewModel.setQuery(it, selectedTags)
                 isModified = true
               },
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/se/icebreakrr/ui/authentication/ProfileCreation.kt
```diff
@@ -280,7 +281,7 @@ fun ProfileCreationScreen(
               },
               height = screenHeight,
               width = screenWidth,
-              textSize = (screenHeight.value * ProfileCreationConstants.TEXT_SIZE_MULTIPLIER).sp)
+              textSize = (screenHeight.value * 0.03).sp)
```
BotondAKovacs, line 284: Nitpick: Magic number

BotondAKovacs (approved):
LGTM

PR #140, opened 2024-11-15
Hotfix/lock screen and lightmode

MaxCastelCast:
This small pr adds some settings to meet the requirments :
-locks the screen horizontaly
-locks the app to light mode


Kapestian (approved):
Sonar cloud doesn't like the fixed screen orientation. As now your fix brings stability to the app I'll approve the changes, but later we should think about landscape and tablet mode

PR #157, opened 2024-11-26
Feature/meeting request v2 2

MaxCastelCast:
Here is the enhanced version of the meeting request system that has the main userflow experience:
-When going in the Inbox page, we can see the profiles that send us a meeting request
-On click, the request is accepted (I lacked the UI to make the choice in the Inbox that will be done by Samuel)
-The user that send the meeting response receives a notification informing him that his meeting request has bee confirmed
-The user that received the meeting request receives a notification with the current location of the meeting request sender

There are a some issues with this implementation, like bugs or lack of features that will be added during the weeks that are ahead of us : 
-The coverage is still not enough
-The coordinates received as a notification is temporary, should appear as a ping on a map later on
-When going in the Notification page, there is a refresh issue, user has to go in there multiple times to be refreshed
-Messages are ignored for now (Don't have the UI yet to display it)
-Can only accept meeting requests (the other case is already taken into account but lack of UI infrastructure to use it)
-Had to addapt some code of the ProfileViewModel to us it
-End to end tests modified because they don't reflect the current Notification screen anymore

https://github.com/user-attachments/assets/b78da0ac-d151-4d2c-ae5a-643883e9bdb4


app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
```diff
@@ -21,9 +21,16 @@ class MeetingRequestViewModel(
 ) : ViewModel() {
 
   var meetingRequestState by mutableStateOf(MeetingRequest())
+  var meetingResponseState by mutableStateOf(MeetingResponse())
+  var meetingConfirmationState by mutableStateOf(MeetingConfirmation())
 
-  private val SEND_MESSAGE_FUNCTION_NAME = "sendMessage"
   private val SEND_MEETING_REQUEST = "sendMeetingRequest"
+  private val SEND_MEETING_RESPONSE = "sendMeetingResponse"
+  private val SEND_MEETING_CONFIRMATION = "sendMeetingConfirmation"
```
SamSteu, line 29: these are constants.

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
```diff
@@ -45,15 +52,21 @@ class MeetingRequestViewModel(
     viewModelScope.launch {}
```
SamSteu, line 52: This `init` block can be removed since it doesn't launch anything

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
SamSteu, line 1: - You can add documentation on these functions.
- You can also replace all the hardcoded strings in this file and put them into const variables

_Question_ : why do you do `viewModel.launch{your code}` in some of your functions?

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
```diff
@@ -45,15 +52,21 @@ class MeetingRequestViewModel(
...
   /**
    * Sets the good token for our message, locally and in the profile's backend
    *
    * @param newToken the current token of the user to which we want to send the message
    */
   fun onRemoteTokenChange(newToken: String) {
     meetingRequestState = meetingRequestState.copy(targetToken = newToken)
-    profilesViewModel.getProfileByUid(MeetingRequestManager.ourUid ?: "null")
-    val currentProfile = profilesViewModel.selectedProfile.value
+    senderToken = newToken
```
SamSteu, line 68: Can't you just do : 
```
fun onRemoteTokenChange(senderToken: String) {
    meetingRequestState = meetingRequestState.copy(targetToken = senderToken)
    ...
```

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
```diff
@@ -177,17 +176,17 @@ class ProfilesRepositoryFirestore(
         }
   }
```
SamSteu, line 178: You can add documentation here to explain what this function does, where and why it is called,

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
```diff
@@ -177,17 +176,17 @@ class ProfilesRepositoryFirestore(
...
       onFailure: (Exception) -> Unit
   ) {
-    if (blockedProfiles.isEmpty()) {
+    if (uidList.isEmpty()) {
       onSuccess(emptyList())
       return
     }
     db.collection(collectionPath)
-        .whereIn("uid", blockedProfiles)
+        .whereIn("uid", uidList)
```
SamSteu, line 189: `uid` can be in a const variable

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -351,6 +355,31 @@ open class ProfilesViewModel(
...
+          _inboxProfiles.value = profileList
+          _loading.value = false
+          _isConnected.value = true
+        },
+        onFailure = { e -> handleError(e) })
+  }
+
+  fun getInboxOfSelfProfile() {
+    val inboxUidList = selfProfile.value?.meetingRequestInbox
+    Log.d("Inside Inbox : ", selfProfile.value?.meetingRequestInbox.toString())
```
SamSteu, line 372: You can get rid of this debug log.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -351,6 +355,31 @@ open class ProfilesViewModel(
...
+    if (inboxUidList != null) {
+      val uidsMessageList = inboxUidList.toList()
+      val uidsList = uidsMessageList.map { it.first }
+      val messageList = uidsMessageList.map { it.second }
+      getInboxUsers(uidsList)
+      _inboxItems.value = _inboxProfiles.value.filterNotNull().zip(messageList).toMap()
+      Log.d("inbox Items : ", _inboxItems.value.toString())
```
SamSteu, line 379: same here.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
SamSteu, line 1: You can add documentation in the functions you added.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
```diff
@@ -41,11 +40,16 @@ private const val PASSED = "Passed"
...
+    meetingRequestViewModel: MeetingRequestViewModel
+) {
   val context = LocalContext.current
-  val cardList = profileViewModel.profiles.collectAsState()
-  val navFunction = { Toast.makeText(context, TOAST_MESSAGE, Toast.LENGTH_SHORT).show() }
-
+  profileViewModel.getSelfProfile()
+  profileViewModel.getInboxOfSelfProfile()
+  val cardList = profileViewModel.inboxItems.collectAsState()
+  Log.d("INBOX VALUE", cardList.value.toString())
```
SamSteu, line 52: You can get rid of this debug log.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
```diff
@@ -73,19 +77,19 @@ fun NotificationScreen(navigationActions: NavigationActions, profileViewModel: P
...
+                    ProfileCard(
+                        p.key,
+                        onclick = {
+                          if (p.key.fcmToken != null) {
+                            meetingRequestViewModel.setMeetingResponse(
+                                p.key.fcmToken!!, MEETING_REQUEST_ACCEPTED, true)
+                            meetingRequestViewModel.sendMeetingResponse()
+                            meetingRequestViewModel.removeFromMeetingRequestInbox(p.key.uid)
+                            profileViewModel.getSelfProfile()
+                            profileViewModel.getInboxOfSelfProfile()
```
SamSteu, line 90: You can add error management here if the `fcmToken` is `null` so that the application doesn't crash if this happens.

app/src/test/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModelTest.kt
```diff
@@ -167,11 +215,92 @@ class MeetingRequestViewModelTest {
 
     val capturedData = argumentCaptor.value as Map<String, Any>
 
+    println(capturedData["targetToken"])
+    println(capturedData["senderUID"])
+    println(capturedData["message"])
```
SamSteu, line 220: You can get rid of these prints.

SamSteu (changes requested):
Overall good PR, you did a lot of work on a task which is difficult to implement. But there are still a few things you need to check.
- Make sure you document each functions that you add or update. Since you are doing backend, some comments inside hard parts of functions can be useful to understand.
- You can get rid of the hardcoded values (mostly strings) inside your code, there are some debugging logs or prints that you can remove too.
- I asked some questions about parts that I didn't understood well on your code.

app/src/androidTest/java/com/github/se/icebreakrr/ui/sections/NotificationTest.kt
```diff
@@ -21,91 +18,104 @@ import com.github.se.icebreakrr.ui.navigation.Route
...
+    }
+
+    val profile = mockProfile()
+    `when`(mockProfilesRepository.getMultipleProfiles(any(), any(), any())).thenAnswer { invocation
+      ->
+      val onSuccessCallback = invocation.getArgument<(List<Profile>) -> Unit>(1)
+      onSuccessCallback(listOf(profile))
+    }
 
     `when`(navigationActions.currentRoute()).thenReturn(Route.NOTIFICATIONS)
```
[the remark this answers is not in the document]
MaxCastelCast: Since those tests are incomplete, i thought about using those values like mockPPRepository, mockAuth and function later on, because they ll be needed but i can do as you ask as a temporary solution that is more clean

app/src/androidTest/java/com/github/se/icebreakrr/ui/sections/NotificationTest.kt
```diff
@@ -21,91 +18,104 @@ import com.github.se.icebreakrr.ui.navigation.Route
...
-        invocation ->
-      val onSuccessCallback = invocation.getArgument<(List<Profile>) -> Unit>(2)
-      onSuccessCallback(listOf(profile))
-      null
-    }
-
-    profilesViewModel.getFilteredProfilesInRadius(GeoPoint(0.0, 0.0), 300)
-    composeTestRule.setContent { NotificationScreen(navigationActions, profilesViewModel) }
-    composeTestRule.onAllNodesWithTag("profileCard").onFirst().assertIsDisplayed()
-    composeTestRule.onAllNodesWithTag("profileCard").assertCountEquals(1)
```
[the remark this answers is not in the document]
MaxCastelCast: Well, when the Inbox List isn't empty, the use ll have displayed some profile cards and i didn't manage to do this test with the new version of the Notification screen, those test can be added at a later point. But those tests were AroundYouTest duplicates so they were of no use

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -246,6 +244,12 @@ fun IcebreakrrNavHost(
         route = Route.AROUND_YOU,
     ) {
       composable(Screen.AROUND_YOU) {
+        profileViewModel.getSelfProfile()
+        profileViewModel.getInboxOfSelfProfile()
+        meetingRequestViewModel?.setInitialValues(
+            MeetingRequestManager.ourFcmToken ?: "null",
+            MeetingRequestManager.ourUid ?: "null",
+            MeetingRequestManager.ourName ?: "null")
```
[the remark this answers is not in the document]
MaxCastelCast: Ok this has been updated at the time i'm writting my comment but the idea is that we need to get the token, uid and name from the SignIn screen (the Manager does that already) and then we need to set the values in the Meeting Request VM as soon as possible. That is why I set those values up in the AroundYou, because it has to be done after the Auth/Sign In step

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -281,7 +285,14 @@ fun IcebreakrrNavHost(
         startDestination = Screen.NOTIFICATIONS,
         route = Route.NOTIFICATIONS,
     ) {
-      composable(Screen.NOTIFICATIONS) { NotificationScreen(navigationActions, profileViewModel) }
+      composable(Screen.NOTIFICATIONS) {
+        if (meetingRequestViewModel != null) {
```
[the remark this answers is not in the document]
MaxCastelCast: It is done here : 
image

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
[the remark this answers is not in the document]
MaxCastelCast: 100% agree with your statement, I ll add the documentation

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
MaxCastelCast: I agree with your remarks, I ll addapt my code to them.

Answer : The one in MeetingRequestViewModel can be removed since it does nothing. I thought it was good practice because i saw it in some corrections in the Bootcamp. But it is true that we don't really use it in our code base

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
MaxCastelCast: True, I ll put them at the begging of the file

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
MaxCastelCast: Yes you are right, i ll remove it

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
MaxCastelCast: Yes this changes the local token, but the token stored in our Profile firebase also needs to be updated. The onRemoteTokenChange function is only called in the MeetingRequestService in the onTokenChange, which is a function that Firebase can trigger at any time. The only purpose of this function is to keep the token in line with Firebase. It shouldn't be used otherwise

MaxCastelCast: You can have more details on this if you read the documentation of FirebaseMessagingService()

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
MaxCastelCast: Yes I can add that !

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesRepositoryFirestore.kt
MaxCastelCast: True, I haven't coded this function but i can set this string as a constant

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
MaxCastelCast: Yes, i ll add documentation

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
MaxCastelCast: Yes I will

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
MaxCastelCast: Yes !

app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
MaxCastelCast: Yes i ll add an error log. But normaly it is impossible that the token is null since for all users, the fcm token is set up at SignIn so the error would have appeared before

MaxCastelCast (commented):
Yes, i ll push the updated code as soon as possible

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
SamSteu: I reviewed your code and you only used `viewModel.launch{your code}` when you have asynchronous operation combined with `await()`, which will delegate that task to a background thread so that the main thread with the UI keeps its responsiveness. You can leave it like that it is, indeed, good practice

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestViewModel.kt
SamSteu: I thought that you were instantiating senderToken here, thank you for the clarifications

SamSteu (approved):
LGTM

SamSteu (approved):
LGTM

PR #160, opened 2024-11-26
feature : report (back-end)

louiscrc:
# Report users

## New implementation of Profile Data Classs
The main change is the add of the `reports` field to the `Profile` data class : it maps a `uid : String` to a `reportType`. 

## Update of the ProfilesViewModel
`reportUser` is a new function : every time one blocks a user X, X's profile.reports is updated, their pair `<reporterUID,reportType>` is added to the map, and the profile is then updated through update Profile

## Update of the repository
The `documentToProfile` has been updated to get the map from Firebase.

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/Profile.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/Profile.kt
@@ -29,7 +29,7 @@ data class Profile(
     val location: GeoPoint? = null,
     val geohash: String? = null,
     var hasBlocked: List<String> = listOf(),
-    var reports : Map<String, reportType> = mapOf(),
+    var reports: Map<String, reportType> = mapOf(),
     val meetingRequestSent: List<String> = listOf(),
     val meetingRequestInbox: Map<String, String> = mapOf()
 ) {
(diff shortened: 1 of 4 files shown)


BotondAKovacs (commented):
Good PR it seems to work. 
Please link it to an issue.
Coverage is fine, but if you have time it would be nice to make it pass the 80% threshold.
Question: Shouldn't we remove from the around you people that have been reported X times?
Question: Shouldn't we add an "unreport" button for testing purposes?

louiscrc:
> Coverage is fine, but if you have time it would be nice to make it pass the 80% threshold.

I've spent 3 hours trying to increase the coverage by testing the ViewModel, but it's impossible because the functions I implemented rely on the selfUser, which is obtained through firebase authentification. Thus I would need to mock all firebase  to make it work...

> Question: Shouldn't we remove from the around you people that have been reported X times?

This is already implemented, in the ViewModel the getFilteredProfilesAround you now contains`(profile.reports.keys.size <= 1)` to remove anyone with too many bans. For modularity, I just pushed a slight modification : `(profile.reports.keys.size <= MAX_REPORTS_BEFORE_BAN)` and `MAX_REPORTS_BEFORE_BAN=2`

> Question: Shouldn't we add an "unreport" button for testing purposes?

I was wondering about doing this at first, but I don't think its necessary. Any tester have an access to firebase and can remove it manually. And coaches can just see that the profile doesn't appear anymore and change their location to see other profiles.

BotondAKovacs (approved):
Your answers make sense to me so LGTM

louiscrc:
For the record, I know this PR isn't huge in terms of ligns. But this is because I did most of the UI alongside the blocking feature in a previous PR (and UI is what takes lines generally). And I didn't implement many tests because as I stated previously, the function uses selfProfile, and testing would require me to mock all the firebase auth (and I don't think it was worth it with already 75% coverage).

PR #161, opened 2024-11-26
Fix : build apk

louiscrc:
- Update of the `Action` to use the Google Maps API key in its environment
- Update of the action triggering : you should manually trigger it (and specify the branch) rather than automatically trigger it on push on the main


BotondAKovacs (approved):
LGTM

PR #174, opened 2024-11-28
Feature/crop image part 2/2

Kapestian:
# Crop Image (2/2)

## Description 
This PR is the last one regarding profile pictures. It brings a whole assortment of fixes and improvements that make choosing, editing and deleting profile pictures perfectly operational.

>[!Note] 
As I am running out of time and energy, I decided that I will take care of the missing tests later, probably during the consolidation phase of the project. 

### Features:
- Every major bugs of part 1 are now fixed, for example:
  - Text changes are not discarded anymore after cropping (most of the new code)
  - No more invalid temporary image used as placeholder
  - The profile upload is now far more optimized (only one upload for image and one for data)
- There's now an option to cancel or delete the image (based on the current edition state)

### State for modifications
In the first version, cropping an image will discard every other information. For this I added new states to the viewmodel and played with them and the isModified variable to keep track of what was changed, enable the discard prompt if needed and avoid useless uploads.

### Better performances
The first version had race condition, duplicate database upload and a lot of mistakes due to last weeks rush. The code is now cleaner, more performant and using the app feels better.
Here is a diagram to understand how the upload works:
image

### Delete Button
When entering the edition screen, the user can now delete it's profile picture if needed. If a user added a profile picture. they may want to cancel and go back to the one they had before entering the edition screen and this possible thanks to the "delete" button adapting to become a "revert" button depending on the changes brought to the picture. 

Before change (possible to delete)
image 

After change (possible to discard and go back to previous picture)
image

## Possible improvements
- Buttons may need to be arranged in another way for usability
- For compatibility, I had to add code waiting for the other screens to use `selfProfile` instead of `selectedProfile`. This must be deleted afterward

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -269,7 +269,10 @@ open class ProfilesViewModel(
    * @param imageData The byte array representing the image data.
    * @param onSuccess A callback function that is invoked when the upload is successful.
    */
-  private fun uploadCurrentUserProfilePicture(imageData: ByteArray, onSuccess: (url: String?) -> Unit) {
+  private fun uploadCurrentUserProfilePicture(
+      imageData: ByteArray,
+      onSuccess: (url: String?) -> Unit
+  ) {
     val userId = _selfProfile.value?.uid ?: throw IllegalStateException("User not logged in")
     ppRepository.uploadProfilePicture(
         userId = userId,
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -103,6 +103,11 @@ open class ProfilesViewModel(
     TO_DELETE
   }
 
+  /**
+   * Updates the profile picture change state.
+   *
+   * @param state The new state of the profile picture change.
+   */
   fun setPictureChangeState(state: ProfilePictureState) {
     _pictureChangeState.value = state
   }
@@ -261,8 +266,8 @@ open class ProfilesViewModel(
   /**
    * Uploads the current user's profile picture to the remote storage system.
    *
-   * @param imageData The byte array of the image file to be uploaded.
-   * @throws IllegalStateException if the user is not logged in.
+   * @param imageData The byte array representing the image data.
+   * @param onSuccess A callback function that is invoked when the upload is successful.
    */
   fun uploadCurrentUserProfilePicture(imageData: ByteArray, onSuccess: (url: String?) -> Unit) {
     val userId = _selfProfile.value?.uid ?: throw IllegalStateException("User not logged in")
@@ -320,6 +325,11 @@ open class ProfilesViewModel(
     _pictureChangeState.value = ProfilePictureState.TO_UPLOAD
   }
 
+  /**
+   * Validates the profile changes and updates both the image and the profile repository.
+   *
+   * @param context The context used to show a Toast message in case of failure.
+   */
   fun validateProfileChanges(context: Context) {
     when (_pictureChangeState.value) {
       ProfilePictureState.TO_UPLOAD ->
@@ -334,7 +344,7 @@ open class ProfilesViewModel(
 
             resetProfileEditionState()
           }
-      ProfilePictureState.TO_DELETE ->
+      TO_DELETE ->
           deleteCurrentUserProfilePicture() {
             val newProfile = _editedCurrentProfile.value?.copy(profilePictureUrl = null)
             updateProfile(newProfile!!)
@@ -352,9 +362,10 @@ open class ProfilesViewModel(
   }
 
   /**
-   * Validates and uploads the profile picture if a temporary profile picture Bitmap exists.
+   * Validates and uploads the profile picture to the remote storage system.
    *
    * @param context The context used to show a Toast message in case of failure.
+   * @param onSuccess A callback function that is invoked when the upload is successful.
    */
   fun validateAndUploadProfilePicture(context: Context, onSuccess: (url: String?) -> Unit) {
     val imageData = bitmapToJpgByteArray(tempProfilePictureBitmap.value ?: return)
@@ -368,7 +379,7 @@ open class ProfilesViewModel(
   /**
    * Deletes the current user's profile picture from the remote storage system.
    *
-   * @throws IllegalStateException if the user is not logged in.
+   * @param onSuccess A callback function that is invoked when the deletion is successful.
    */
   fun deleteCurrentUserProfilePicture(onSuccess: () -> Unit) {
     ppRepository.deleteProfilePicture(
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -8,6 +8,8 @@ import android.util.Log
 import android.widget.Toast
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
+import com.github.se.icebreakrr.model.profile.ProfilesViewModel.ProfilePictureState.TO_DELETE
+import com.github.se.icebreakrr.model.profile.ProfilesViewModel.ProfilePictureState.UNCHANGED
 import com.github.se.icebreakrr.ui.sections.DEFAULT_RADIUS
 import com.github.se.icebreakrr.ui.sections.DEFAULT_USER_LATITUDE
 import com.github.se.icebreakrr.ui.sections.DEFAULT_USER_LONGITUDE
--- a/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
@@ -269,7 +269,7 @@ open class ProfilesViewModel(
    * @param imageData The byte array representing the image data.
    * @param onSuccess A callback function that is invoked when the upload is successful.
    */
-  fun uploadCurrentUserProfilePicture(imageData: ByteArray, onSuccess: (url: String?) -> Unit) {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -269,34 +284,84 @@ open class ProfilesViewModel(
...
+  fun validateProfileChanges(context: Context) {
+    when (_pictureChangeState.value) {
+      ProfilePictureState.TO_UPLOAD ->
+          validateAndUploadProfilePicture(context) { url ->
+            val newProfile = _editedCurrentProfile.value!!.copy(profilePictureUrl = url)
+            updateProfile(newProfile)
+            _selfProfile.value = newProfile
+            // todo: "_selectedProfile.value = newProfile" needs to be removed when all the other
+            // personal profile views are updated to use the selfProfile instead of the
+            // selectedProfile
```
MaxCastelCast, line 342: It isn't recommended to leave commented code (I have lost points in my M2 individual because i had left commented code). If you want to mark a todo, it is better to write "//TODO:" as it is the current standard used in our codebase

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -269,34 +284,84 @@ open class ProfilesViewModel(
...
+            _selectedProfile.value = newProfile
```
MaxCastelCast, line 343: Question : why do we have to change the selected profile value ? I don't see the use here because we change the profile picture of our profile and the selected profile is used for the AroundYou page or the Notification page to see the profile of other people. Maybe I missunderstood your code

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -269,34 +284,84 @@ open class ProfilesViewModel(
...
+        resetProfileEditionState()
+      }
+    }
   }
 
   /**
    * Validates and uploads the profile picture if a temporary profile picture Bitmap exists.
    *
    * @param context The context used to show a Toast message in case of failure.
    */
```
MaxCastelCast: There is a lack of documentation here, the parameter onSucces isn't described

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -305,15 +370,10 @@ open class ProfilesViewModel(
    *
    * @throws IllegalStateException if the user is not logged in.
    */
```
MaxCastelCast: No documentation here for the onSucces variable

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -244,17 +264,12 @@ open class ProfilesViewModel(
    * @param imageData The byte array of the image file to be uploaded.
    * @throws IllegalStateException if the user is not logged in.
    */
-  fun uploadCurrentUserProfilePicture(imageData: ByteArray) {
-    val userId = selectedProfile.value?.uid ?: throw IllegalStateException("User not logged in")
+  fun uploadCurrentUserProfilePicture(imageData: ByteArray, onSuccess: (url: String?) -> Unit) {
```
MaxCastelCast, line 272: onSuccess isn't described in the documentation

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -87,6 +96,17 @@ open class ProfilesViewModel(
...
+  /** Represents the state of a profile picture change. */
+  enum class ProfilePictureState {
+    UNCHANGED,
+    TO_UPLOAD,
+    TO_DELETE
+  }
+
+  fun setPictureChangeState(state: ProfilePictureState) {
+    _pictureChangeState.value = state
+  }
```
MaxCastelCast, line 113: Even if it seems obvious what those lines do, a description of this function could be nice to keep code uniformity

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
```diff
@@ -269,34 +284,84 @@ open class ProfilesViewModel(
...
   fun processCroppedImage(bitmap: Bitmap) {
     _tempProfilePictureBitmap.value = compressImageIfTooLarge(bitmap)
+    _pictureChangeState.value = ProfilePictureState.TO_UPLOAD
+  }
+
+  fun validateProfileChanges(context: Context) {
+    when (_pictureChangeState.value) {
+      ProfilePictureState.TO_UPLOAD ->
+          validateAndUploadProfilePicture(context) { url ->
+            val newProfile = _editedCurrentProfile.value!!.copy(profilePictureUrl = url)
```
MaxCastelCast, line 337: Small detail / Question : If the function "validateProfileChanges(...)" is called before "saveEditedProfile" at the beginning, then "_editedCurrentProfile.value" can be null, so is it safe to set "!!" here ?

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
```diff
@@ -152,14 +160,28 @@ fun ProfileEditingScreen(
...
+                    onDeletion = {
+                      when (pictureChangeState) {
+                        UNCHANGED -> { // delete current profile picture
+                          if (user.profilePictureUrl != null) {
+                            profilesViewModel.setPictureChangeState(TO_DELETE)
+                            isModified = true
+                          }
+                        }
+                        // else cancel changes
+                        else -> profilesViewModel.setPictureChangeState(UNCHANGED)
```
MaxCastelCast, line 182: Question : Why using a "when" here is you have only two conditions : could you use just "if else" or is there an other reason behind this ?

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfilePictureSelector.kt
```diff
@@ -43,9 +48,11 @@ private val ICON_PADDING = 5.dp
 fun ProfilePictureSelector(
     url: String?,
     localBitmap: Bitmap?,
+    pictureChangeState: ProfilesViewModel.ProfilePictureState,
     size: Dp,
     onSelectionSuccess: (Uri) -> Unit,
-    onSelectionFailure: () -> Unit
+    onSelectionFailure: () -> Unit,
+    onDeletion: () -> Unit,
```
MaxCastelCast, line 55: Lack of documentation for those new parameters

app/src/main/java/com/github/se/icebreakrr/ui/profile/profilePictureEdit.kt
```diff
@@ -64,6 +65,13 @@ fun ImageCropperScreen(
           navigationActions.goBack()
         })
   }
+
+  BackHandler {
+    // for now makes impossible to quit the screen without cropping (as this may discard a
+    // valid temporary profile picture). A possible improvement for the future could be to add
+    // a slot to save the not cropped image so it doesn't discard the valid actual
+    // temporary image.
```
MaxCastelCast, line 73: This is a very good idea ! Maybe we should add it in the scrum board to make this feature available for the following weeks

MaxCastelCast (changes requested):
The code seems globally very nice, there is juste a lack of documentations and there are other parts of the code that are unclear. The requested changes are quick to do so it should be ok

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
Kapestian: That's exactly what I have done. This is not code, this is a comment tagged with `//todo:` to be spotted later on. I also opened an issue to remember to replace `currentProfile` with `selfProfile` in the other user screens.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
Kapestian: The reason is written in the comment just above the assignment :) 
Old code in settings and `ProfileView` currently still uses `selectedProfile` to display the current users, which leads to several issues after leaving the edition screen (changes not displayed). I Fixed it in `ProfileEdit` but I didn't want to modify every other screen as it was not my task and I could have disturbed someone else's job. That's why I made this here and that I tagged it to be removed later.

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
Kapestian: OUCH... you didn't read my last message on discord saying to wait 5m for me to update the outdated doc. You started reviewing 30s before the push was made leading you to review an outdated code. That one's on me, I should have been more clear. I'll marke the other comment's with an emoji

app/src/main/java/com/github/se/icebreakrr/model/profile/ProfilesViewModel.kt
Kapestian: I understand your concern. There is only one place this function is called, and I used saveEditedProfile before. Now that I think of it I could have passed the changes as parameter and avoid this saving call from `ProfileEdit`, but as I am now well into test development, I don't want to break everything

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfileEdit.kt
Kapestian: I always use the `when` syntax with enums as I find it more readable and more scalable (if I have to add more states to the enum, this code will be easy to update and still readable.)

app/src/main/java/com/github/se/icebreakrr/ui/profile/ProfilePictureSelector.kt
Kapestian: That one was missed even after my documentation update, thanks!

MaxCastelCast (approved):
Looks good to me !

PR #181, opened 2024-12-02
feat: add a map location selector

SamSteu:
This PR introduces a new screen to chose the location of the meeting and send a message to precise it. It also refactors some code to add `onComplete()` callback functions on some view model functions to avoid racing conditions. Some logic has been added for the badges (to indicate notifications) in the notification screen and the bottom navigation bar

 hese are the principal files I have changed : 

- `Profile` : added `meetingRequestPendingLocation`, which contains all the meetings where you need to select the location and 
     `meetingRequestChosenLocalisation` which contains the information of the meeting that is ready, so uid, location, and a 
     message.
- `TopLevelDestination` : add a new screen and route for the new map
- `ProfileRepositoryFirestore` : I needed to modify `documentToProfile` to take into account the two new attributes of the `Profile`
- `ProfileViewModel` : Added two functions to fetch the pending profiles and the chosen locations. I also added some 
    `onComplete` callbacks to avoid racing conditions and added other helper functions related to that
- `MainActivity` : add `LocationSelectorMapScreen` to the navigation graph
- `MeetingRequestService` : Modified to to fit our new messaging system.
- `MeetingRequestViewModel` : add functions to deal with the pending locations and chosen locations. I also modified 
     `updateProfile` to have a `onComplete` callback
- `HeatMap`, `AroundYou`, `Settings`, `AlreadyMet`, `UnblockProfile` : modifier parameters in the `BottomNavigationBar` to update the notification badge on the heatmap (the number of meeting that are ready with a localisation and a message) and on the inbox
- `LocationSelectorMap` : added this composable to be able to select a meeting point in a range of 500m and send a message to the other user
- `BottomNavigationMenu` : add logic for the badges
- `Notifications` : add logic to handle the badges
- `TopBar` : make it adaptable to the size of the text
- `LocationMapSelectorTest` : File that tests the `LocationSelectorMapScreen`
- `NotificationTest` : Add tests to check if inbox, pending, and sent get appropriately updated
- `ProfilesViewModelTest` : added tests for the new functions

https://github.com/user-attachments/assets/77ec1a4b-09e2-4573-8786-4ee58af174a0

It seems like my SonarCloud still doesn't take into account my new tests because it tells me that 0% of the code has been covered on the file `LocationSelectorMap`, but jacoco doesn't tell me the same thing as you can see : 

image

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
@@ -35,60 +35,53 @@ class MeetingRequestService : FirebaseMessagingService() {
     val senderUid = remoteMessage.data["senderUID"] ?: "null"
     val message = remoteMessage.data["message"] ?: "null"
     val title = remoteMessage.data["title"] ?: "null"
+    val senderName = remoteMessage.data["senderName"] ?: "null"
 
     when (title) {
       "MEETING REQUEST" -> {
-        val name = remoteMessage.data["senderName"] ?: "null"
         MeetingRequestManager.meetingRequestViewModel?.addToMeetingRequestInbox(
             senderUid, message) {
               MeetingRequestManager.meetingRequestViewModel?.updateInboxOfMessages() {}
             }
-        showNotification(MSG_REQUEST, "from : $name")
+        showNotification(MSG_REQUEST, "from : $senderName")
       }
       "MEETING RESPONSE" -> {
-        val name = remoteMessage.data["senderName"] ?: "null"
         val accepted = remoteMessage.data["accepted"]?.toBoolean() ?: false
-        val senderToken = remoteMessage.data["senderToken"] ?: "null"
 
         MeetingRequestManager.meetingRequestViewModel?.removeFromMeetingRequestSent(senderUid) {
           MeetingRequestManager.meetingRequestViewModel?.updateInboxOfMessages() {}
         }
         if (accepted) {
-          showNotification(name + MSG_RESPONSE_ACCEPTED, "")
+          showNotification(senderName + MSG_RESPONSE_ACCEPTED, "")
           MeetingRequestManager.meetingRequestViewModel?.addPendingLocation(senderUid) {}
         } else {
-          showNotification(name + MSG_RESPONSE_REJECTED, "")
+          showNotification(senderName + MSG_RESPONSE_REJECTED, "")
         }
       }
       "MEETING CONFIRMATION" -> {
-        val name = remoteMessage.data["senderName"] ?: "null"
-        val mewSenderUid = remoteMessage.data["senderUID"] ?: "null"
         val locationString = remoteMessage.data["location"] ?: "null"
-        val newMessage = remoteMessage.data["message"] ?: ""
         val latitudeString = locationString.split(", ")[0]
         val longitudeString = locationString.split(", ")[1]
         MeetingRequestManager.meetingRequestViewModel?.confirmMeetingLocation(
-            senderUid,
-            Pair(newMessage, Pair(latitudeString.toDouble(), longitudeString.toDouble())))
-        showNotification(name + MSG_CONFIRMATION, MSG_CONFIRMATION_INFO)
+            senderUid, Pair(message, Pair(latitudeString.toDouble(), longitudeString.toDouble()))) {
+              Log.e("MeetingRequestService", "error when confirmMeetingLocation : ${it.message}")
+            }
+        showNotification(senderName + MSG_CONFIRMATION, MSG_CONFIRMATION_INFO)
       }
       "MEETING CANCELLATION" -> {
-        val name = remoteMessage.data["senderName"] ?: "null"
-        Log.d("CANCELLATION REASON", message)
         val stringReason =
             when (message) {
               "distance" -> DISTANCE_REASON_CANCELLATION
               else -> DEFAULT_REASON_CANCELLATION
             }
-        showNotification("Cancelled meeting with $name", stringReason)
+        showNotification("Cancelled meeting with $senderName", stringReason)
         MeetingRequestManager.meetingRequestViewModel?.removeFromMeetingRequestInbox(senderUid)
         MeetingRequestManager.meetingRequestViewModel?.removeFromMeetingRequestSent(senderUid) {}
       }
       "ENGAGEMENT NOTIFICATION" -> {
-        val name = remoteMessage.data["senderName"] ?: "null"
         showNotification(
             "A person with similar interests is close by !",
-            "The user $name has the common tag : $message")
+            "The user $senderName has the common tag : $message")
       }
     }
   }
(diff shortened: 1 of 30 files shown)


app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
```diff
@@ -39,35 +39,38 @@ class MeetingRequestService : FirebaseMessagingService() {
     when (title) {
       "MEETING REQUEST" -> {
```
JanStaszewiczEPFL, line 40: Great use of when cases to make clean and well-organized code !

when (title) {
    "MEETING REQUEST" -> { /* Logic here */ }
    "MEETING RESPONSE" -> { /* Logic here */ }
    "MEETING CONFIRMATION" -> { /* Logic here */ }
    "MEETING CANCELLATION" -> { /* Logic here */ }
    "ENGAGEMENT NOTIFICATION" -> { /* Logic here */ }
}

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
```diff
@@ -121,7 +124,6 @@ class MeetingRequestService : FirebaseMessagingService() {
             .setContentText(message)
             .setPriority(NotificationCompat.PRIORITY_DEFAULT)
             .setAutoCancel(true)
-
     notificationManager.notify(NOTIFICATION_ID, notificationBuilder.build())
```
JanStaszewiczEPFL, line 127: NOTIFICATION_ID is hardcoded to 0, shouldn't notifications have unique ids to ensure they don't overwrite each other?

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
```diff
@@ -39,35 +39,38 @@ class MeetingRequestService : FirebaseMessagingService() {
...
+          MeetingRequestManager.meetingRequestViewModel?.addPendingLocation(senderUid) {}
         } else {
           showNotification(name + MSG_RESPONSE_REJECTED, "")
-          MeetingRequestManager.meetingRequestViewModel?.setMeetingConfirmation(
-              targetToken = senderToken,
-              newMessage = "The meeting with ${MeetingRequestManager.ourName} is cancelled !")
         }
       }
       "MEETING CONFIRMATION" -> {
         val name = remoteMessage.data["senderName"] ?: "null"
```
JanStaszewiczEPFL, line 64: You get the smalls values from remoteMessage.data several times, you can extract them once into variables : val senderUid = remoteMessage.data["senderUID"] ?: "null"
val senderName = remoteMessage.data["senderName"] ?: "null"
val message = remoteMessage.data["message"] ?: "null"

app/src/main/java/com/github/se/icebreakrr/ui/map/LocationSelectorMap.kt
```diff
@@ -0,0 +1,201 @@
...
+          IconButton(
+              onClick = {
+                if (mapLoaded || isTesting) {
+                  meetingRequestViewModel.confirmMeetingLocation(
+                      profileId!!,
+                      Pair(
+                          stringQuery,
+                          Pair(
+                              markerState?.position?.latitude!!,
+                              markerState?.position?.longitude!!)))
```
JanStaszewiczEPFL, line 143: "!!" could lead to crashes if the values are null, try adding more none null logic to the code

app/src/main/java/com/github/se/icebreakrr/ui/map/LocationSelectorMap.kt
```diff
@@ -0,0 +1,201 @@
...
+                  meetingRequestViewModel.setMeetingConfirmation(
+                      profile.value?.fcmToken!!,
+                      markerState?.position?.latitude!!.toString() +
+                          ", " +
+                          markerState?.position?.longitude!!.toString(),
+                      stringQuery)
+                  meetingRequestViewModel.sendMeetingConfirmation()
+                  navigationActions.navigateTo(Route.HEAT_MAP)
+                }
```
JanStaszewiczEPFL, line 152: There are no feedback provided for errors during meeting confirmation or ViewModel operations. For instance a toast could be added if something goes wrong with the view model.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
```diff
@@ -99,57 +104,88 @@ fun NotificationScreen(
...
+                      Column(verticalArrangement = Arrangement.spacedBy(CARD_SPACING)) {
+                        pendingLocation.value.forEach { p ->
+                          ProfileCard(
+                              profile = p,
+                              onclick = {
+                                navigationActions.navigateTo(
+                                    Screen.MAP_MEETING_LOCATION_SCREEN + "?userId=${p.uid}")
+                              },
+                              greyedOut = false)
+                        }
```
JanStaszewiczEPFL, line 184: The LazyColumn blocks for INBOX, SENT, and CHOOSE_LOCATION options share similar code so you could extract the common logic into a reusable composable function.

app/src/main/java/com/github/se/icebreakrr/ui/sections/shared/TopBar.kt
```diff
@@ -68,9 +68,10 @@ fun TopBar(s: String, needBackButton: Boolean = false, backButtonOnClick: () ->
                       modifier = Modifier.size(BACK_BUTTON_ICON_SIZE))
                 }
           }
+          val topBarSize = if (s.length < 15) TOP_BAR_TEXT_SIZE else 25.sp
```
JanStaszewiczEPFL, line 71: Why 15? It's a magic number!

JanStaszewiczEPFL (changes requested):
This is a big PR, there is a lot of code and it would maybe be safer to have an additional reviewer. I requested a few changes, otherwise great PR and overall clean code.

app/src/main/java/com/github/se/icebreakrr/ui/sections/Notifications.kt
SamSteu: That's a good refactoring, thank you. It has been done

app/src/main/java/com/github/se/icebreakrr/ui/sections/shared/TopBar.kt
SamSteu: True! It is because over 15 character, the text uses 2 lines in the top app bar. I've put these into const variables

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
JanStaszewiczEPFL: Okay we will see!

app/src/main/java/com/github/se/icebreakrr/ui/map/LocationSelectorMap.kt
JanStaszewiczEPFL: Okay I see now thanks

JanStaszewiczEPFL (approved):

JanStaszewiczEPFL (approved):
LGTM

PR #188, opened 2024-12-04
Feature/engagement notification

BotondAKovacs:
### Summary of changes:

- Create a new EngagementNotificationManager.kt that defines the functions needed for this notification
- checks if there are any users with at least one matching tag in the users' radius
- if there is it sends a notification to incite the user to come on the app to meet this person
- the notification is only sent if the user doesn't have the app in foreground ( this is deactivated for now, as the location doesn't update if the user doesn't have the app in the foreground, the feature wouldn't work at all with this)
- the number of notifications is capped at 1 every 5 minutes but this can easily be changed
- updated the Around You to launch the monitoring of potential notifications to send
- we track when was the last time a user sent a notification to another user so that users aren't spammed

Screenshot 2024-12-05 at 21 22 45

Testing: Not yet done as I haven't figured out how to test notifications yet.

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
@@ -1,6 +1,5 @@
 package com.github.se.icebreakrr.model.notification
 
-import android.app.ActivityManager
 import android.content.Context
 import android.util.Log
 import androidx.annotation.VisibleForTesting
@@ -69,21 +68,16 @@ class EngagementNotificationManager(
     notificationJob = null
   }
 
-  private fun isAppInForeground(): Boolean {
-    val activityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager
-    val appProcesses = activityManager.runningAppProcesses ?: return false
-    val packageName = context.packageName
 
-    for (appProcess in appProcesses) {
-      if (appProcess.importance == ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND &&
-          appProcess.processName == packageName) {
-        return true
-      }
-    }
-    return false
-  }
-
+    /**
+     * Checks for nearby users with common tags and processes them for potential engagement notifications.
+     *
+     * This function retrieves the user's profile and location, then uses the filter settings to find
+     * nearby profiles within the specified radius. It launches a coroutine to collect these profiles
+     * and processes them if the user is discoverable.
+     */
   private fun checkNearbyUsersForCommonTags() {
+      profilesViewModel.getSelfProfile()
     val selfProfile = profilesViewModel.selfProfile.value ?: return
     val selfLocation = selfProfile.location ?: return
 
@@ -100,19 +94,28 @@ class EngagementNotificationManager(
       if (!appDataStore.isDiscoverable.first()) return@launch
 
       profilesViewModel.filteredProfiles.collectLatest { nearbyProfiles ->
-        // Only send notifications if the app is not in foreground
-        if (!isAppInForeground()) {
-          processNearbyProfiles(selfProfile, nearbyProfiles)
-        }
+        // Process all nearby profiles
+        processNearbyProfiles(selfProfile, nearbyProfiles)
       }
     }
   }
 
+    /**
+     * Processes a list of nearby profiles to find common tags and send engagement notifications.
+     *
+     * @param selfProfile The user's own profile containing their tags.
+     * @param nearbyProfiles A list of profiles that are within the user's selected radius.
+     *
+     * This function filters out the user's own profile from the list, then iterates over the remaining
+     * profiles to find common tags. If common tags are found, it sends a notification to the nearby
+     * user using the first common tag.
+     */
   private fun processNearbyProfiles(selfProfile: Profile, nearbyProfiles: List<Profile>) {
     val selfTags = selfProfile.tags
 
-    for (nearbyProfile in nearbyProfiles) {
-      if (nearbyProfile.uid == selfProfile.uid) continue // Skip self
+      val newListMinusSelf = nearbyProfiles.filter{it != selfProfile}
+
+    for (nearbyProfile in newListMinusSelf) {
 
       // Find common tags
       val commonTags = selfTags.intersect(nearbyProfile.tags.toSet())
@@ -121,13 +124,11 @@ class EngagementNotificationManager(
         // Send notification for the first common tag
         val commonTag = commonTags.first()
         try {
-          // Send notification to the other user
+          // Send notification to other users - receiving device will check if it should show it
           meetingRequestViewModel.engagementNotification(
-              targetToken = nearbyProfile.fcmToken ?: continue, tag = commonTag)
-
-          // Send notification to self as well
-          meetingRequestViewModel.engagementNotification(
-              targetToken = selfProfile.fcmToken ?: continue, tag = commonTag)
+              targetToken = nearbyProfile.fcmToken ?: "null",
+              tag = commonTag
+          )
         } catch (e: Exception) {
           Log.e("EngagementNotification", "Failed to send notification", e)
         }
--- a/app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
@@ -8,7 +8,6 @@ import com.github.se.icebreakrr.model.filter.FilterViewModel
 import com.github.se.icebreakrr.model.message.MeetingRequestViewModel
 import com.github.se.icebreakrr.model.profile.Profile
 import com.github.se.icebreakrr.model.profile.ProfilesViewModel
-import com.github.se.icebreakrr.ui.sections.ALLOWED_VALUES
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -20,7 +19,6 @@ import kotlinx.coroutines.launch
 // This file was written with the help of Cursor AI
 
 private const val CHECK_INTERVAL = 5 * 60 * 1_000_000L // 5 minutes
-private val RADIUS = ALLOWED_VALUES[0]
 private const val NOTIFICATION_COOLDOWN = 4 * 60 * 60 * 1000L // 4 hours in milliseconds
 
 /**
--- a/app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
@@ -4,10 +4,10 @@ import android.app.ActivityManager
 import android.content.Context
 import android.util.Log
 import com.github.se.icebreakrr.data.AppDataStore
+import com.github.se.icebreakrr.model.filter.FilterViewModel
 import com.github.se.icebreakrr.model.message.MeetingRequestViewModel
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+    val packageName = context.packageName
+
+    for (appProcess in appProcesses) {
+      if (appProcess.importance == ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND &&
+          appProcess.processName == packageName) {
+        return true
+      }
+    }
+    return false
+  }
```
MaxCastelCast, line 84: This function is very nice ! It can be useful in a lot of other classes, like in the MeetingRequestViewModel or in the MeetingRequestService. Have two questions : 
1) Could you add a description for this function please ?
2) Could you make this function more accessible anyware in the code ? By removing the private and putting it in a class were any class can call it ?

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+    stopMonitoring() // Stop any existing monitoring
+
+    notificationJob =
+        scope.launch {
+          while (true) {
+            checkNearbyUsersForCommonTags()
+            delay(CHECK_INTERVAL)
+          }
+        }
+  }
```
MaxCastelCast, line 64: I see what the notificationJob doing, it calls every 5 minutes the function to checkNerbyUsers, but when i used delay(...) last time around, Valentin had highly criticized me because it seems like a temporary fix than a proper implementation. Would there be an other way to do this ? But if you think this is clean code, this is fine, juste be sure it is the right way.

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+          appProcess.processName == packageName) {
+        return true
+      }
+    }
+    return false
+  }
+
+  private fun checkNearbyUsersForCommonTags() {
+    val selfProfile = profilesViewModel.selfProfile.value ?: return
+    val selfLocation = selfProfile.location ?: return
```
MaxCastelCast, line 88: Shouldn't you call getSelfProfile to get the updated version ofthe selfProfile ? Because maybe the user changed his profile tags and the changed have been saved in the Firebase, but not locally. So i think a call to refresh the self profile is necessary here

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    // Use the user's selected radius from FilterViewModel
+    profilesViewModel.getFilteredProfilesInRadius(
+        center = selfLocation,
+        radiusInMeters = filterViewModel.selectedRadius.value,
+        genders = filterViewModel.selectedGenders.value,
+        ageRange = filterViewModel.ageRange.value)
```
MaxCastelCast, line 95: Very nice use of Arthur's code ! 
Question of a hypothetical senario : Alice sets here radius at 500m and Bob at 100m, does it make sense that therefore Alice ll receive a lot more engagement notification than Bob because she set a bigger radiius ? If Alice and Bob are at 300m and both have the #BBQRibs , then Alice ll be engaged and not Bob. Is this really the intended use case ? It is food for thoughts and could be discussed next week, it isn't urgent.

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+        }
+      }
+    }
+  }
+
+  private fun processNearbyProfiles(selfProfile: Profile, nearbyProfiles: List<Profile>) {
+    val selfTags = selfProfile.tags
+
+    for (nearbyProfile in nearbyProfiles) {
+      if (nearbyProfile.uid == selfProfile.uid) continue // Skip self
```
MaxCastelCast, line 115: Wouldn't here be a more elegant solution ? like using the Filter function of nerbyProfiles before calling the for loop ? Would make the code more readable and optimize a little bit the code. You could do something like "val newList = nearbyProfiles.filter{it != selfProfile}

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+          // Send notification to the other user
+          meetingRequestViewModel.engagementNotification(
+              targetToken = nearbyProfile.fcmToken ?: continue, tag = commonTag)
+
+          // Send notification to self as well
+          meetingRequestViewModel.engagementNotification(
+              targetToken = selfProfile.fcmToken ?: continue, tag = commonTag)
+        } catch (e: Exception) {
+          Log.e("EngagementNotification", "Failed to send notification", e)
+        }
```
MaxCastelCast, line 133: This part is a good use of my code of the MeetingRequestViewModel, well done ! Just one question tho :
1) Maybe it isn't very clear in my messaging implementation but normally a Profile ll always have an FCM token, because when Signing in, the FCM token is directly added. So i think the "continue" can be replaced by strings like ""null"", like i did in my implementation. What do you think ? The advantage of "null" is that it would trigger errors so it could help for debugging and bad behavior. But the use of "continue " is also correct, but it avoids the errors

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+  }
+
+  // Add this method for testing purposes
+  @VisibleForTesting fun isMonitoring(): Boolean = notificationJob?.isActive == true
```
MaxCastelCast, line 139: This is very clever ! I didn't even now that you can make functions only visible for testing ! I think using this kind of function for testing in the futur

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+      profilesViewModel.filteredProfiles.collectLatest { nearbyProfiles ->
+        // Only send notifications if the app is not in foreground
+        if (!isAppInForeground()) {
+          processNearbyProfiles(selfProfile, nearbyProfiles)
+        }
+      }
+    }
+  }
+
+  private fun processNearbyProfiles(selfProfile: Profile, nearbyProfiles: List<Profile>) {
```
MaxCastelCast, line 111: Lack of documentation for this function

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -0,0 +1,140 @@
...
+    for (appProcess in appProcesses) {
+      if (appProcess.importance == ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND &&
+          appProcess.processName == packageName) {
+        return true
+      }
+    }
+    return false
+  }
+
+  private fun checkNearbyUsersForCommonTags() {
```
MaxCastelCast, line 86: lack of documentation for this function

app/src/test/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManagerTest.kt
```diff
@@ -0,0 +1,97 @@
...
+  @Test
+  fun `test monitoring starts and stops correctly`() {
+    engagementManager.startMonitoring()
+    // Verify monitoring started
+    assert(engagementManager.isMonitoring())
+
+    engagementManager.stopMonitoring()
+    // Verify monitoring stopped
+    assert(!engagementManager.isMonitoring())
+  }
```
MaxCastelCast, line 96: This is a good start for the testing process. I also struggled with testing functions linked with notifications sending. Maybe you could try adding a test that tests if the function engagementNotification is called ? You can take inspiration from the MeetingRequestViewModelTest

MaxCastelCast (changes requested):
Very nice implementation overall ! There is just a lack of documentation in some places and others were i have recommendation/questions. They are small fixes anyway, those can be done quickly.
And after testing on my computer, i see that the feature doesn't work a 100% correctly, so making it work is priority n°1

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
BotondAKovacs: I think we can put it in MeetingRequestService and leave it private, as that file can by itself choose what notifications to show and what not to show. Do you agree? Or maybe I forgot another use case that would justify making it public?

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
BotondAKovacs: I'll look for a better way.

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
BotondAKovacs: You're right, good catch!

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
BotondAKovacs: That's a good question, how do we handle this in around you? Does Alice see Bob but Bob doesn't Alice? Does it auto cancel Alice's meeting request for example?

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
BotondAKovacs: you're right i'll change it

BotondAKovacs:
> And after testing on my computer, i see that the feature doesn't work a 100% correctly, so making it work is priority n°1. When i use the code on my computer and i try to test the feature, i do indeed get a notification, but not the right one !

That's weird because it used to work, I may have broken it in the mean time. Let me fix that.

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
MaxCastelCast: Ok that seems to be a good idea !

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
MaxCastelCast: I think that whatever the radius you choose for the filter, all users should receive the same amount of engagement notifications. And since all users can see the users  around 100m (because minRadius = 100m) then you should set your radiusInMeters to 100 i think. The inconvinent part is that it would limite the reach of the feature you implemented but at least it wouldn't harasse users and keep it consistent (same for all users)

BotondAKovacs: I'll set the radius to the minimum radius set for the filters for now, but I think we'll have to review this in the future and make a consistent decision for the app overall

MaxCastelCast (approved):
Looks good to me !

PR #200, opened 2024-12-05
Test/messaging increase coverage

MaxCastelCast:
Increase coverage for the following classes :
-MeetingRequestViewModel
-MeetingRequestService

I also refactored a bit MeetingRequestService to make it more easily testable


BotondAKovacs (approved):
LGTM

PR #254, opened 2024-12-13
fix: display correct string values in notification

arthur-mrgt:
- Replaced R.string.resource_id.toString() with getString(resource_id) to ensure proper retrieval of string values in the notification's title and content text.


SamSteu (approved):
LGTM

PR #276, opened 2024-12-16
Feature/enhanceengagementandfixoffline

BotondAKovacs:
### Summary:
- fix offline mode in around you
- make notifications clickable
- reinstate sign in screen, remove again sign in screen
- save in data store the time a notification was last sent to a user
- put different IDs for different types of notifications so that they dont override each other
- put monitoring of notification logic back in around you so that we cans skip sign in screen

### DIFF
--- a/app/src/androidTest/java/com/github/se/di/module/TestModule.kt
+++ b/app/src/androidTest/java/com/github/se/di/module/TestModule.kt
@@ -20,6 +20,7 @@ import com.google.firebase.firestore.FirebaseFirestore
 import com.google.firebase.firestore.GeoPoint
 import com.google.firebase.firestore.Query
 import com.google.firebase.firestore.QuerySnapshot
+import com.google.firebase.firestore.Source
 import dagger.Module
 import dagger.Provides
 import dagger.hilt.components.SingletonComponent
@@ -158,7 +159,7 @@ object MockFirebaseFirestoreModule {
         .thenReturn(mockProfilesQuery)
     `when`(mockProfilesQuery.whereLessThanOrEqualTo(anyString(), anyString()))
         .thenReturn(mockProfilesQuery)
-    `when`(mockProfilesQuery.get()).thenReturn(mockTaskQuerySnapshotProfiles)
+    `when`(mockProfilesQuery.get(Source.SERVER)).thenReturn(mockTaskQuerySnapshotProfiles)
     doAnswer { invocation ->
           val listener = invocation.arguments[0] as OnSuccessListener<QuerySnapshot>
           listener.onSuccess(mockQuerySnapshotProfiles)
--- a/app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
@@ -372,7 +372,9 @@ fun IcebreakrrApp(
           tagsViewModel = tagsViewModel,
           permissionManager = permissionManager)
 
-  val startDestination = if (isTesting) Route.AROUND_YOU else Route.AUTH
+  val startDestination =
+      if (isTesting) Route.AROUND_YOU
+      else (if (currentUser != null) Route.AROUND_YOU else Route.AUTH)
 
   IcebreakrrNavHost(
       profileViewModel,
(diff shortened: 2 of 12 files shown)


app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
```diff
@@ -372,9 +372,7 @@ fun IcebreakrrApp(
           tagsViewModel = tagsViewModel,
           permissionManager = permissionManager)
 
-  val startDestination =
-      if (isTesting) Route.AROUND_YOU
-      else (if (currentUser != null) Route.AROUND_YOU else Route.AUTH)
+  val startDestination = if (isTesting) Route.AROUND_YOU else Route.AUTH
```
MaxCastelCast, line 375: Here it would be better to keep the previous code. It is a cool feature to skip the login screen when we have already signed in and that we are re-entering the app

app/src/main/java/com/github/se/icebreakrr/data/AppDataStore.kt
```diff
@@ -127,4 +137,14 @@ open class AppDataStore(private val dataStore: DataStore<Preferences>) {
...
+
+  suspend fun saveNotificationTime(uid: String, timestamp: Long) {
+    putPreference(longPreferencesKey(NOTIFICATION_TIME_PREFIX + uid), timestamp)
+  }
+
+  suspend fun clearNotificationTime(uid: String) {
+    dataStore.edit { preferences ->
+      preferences.remove(longPreferencesKey(NOTIFICATION_TIME_PREFIX + uid))
+    }
+  }
```
MaxCastelCast, line 149: Lack of documentation for those two functions, it isn't clear enough what they are doing

app/src/main/java/com/github/se/icebreakrr/model/notification/EngagementNotificationManager.kt
```diff
@@ -206,6 +215,12 @@ class EngagementNotificationManager(
     }
   }
 
+  private suspend fun updateLastNotificationTime(uid: String) {
+    val currentTime = System.currentTimeMillis()
+    lastNotificationTimes[uid] = currentTime
+    appDataStore.saveNotificationTime(uid, currentTime)
+  }
+
```
MaxCastelCast, line 223: Lack of documentation for this function

app/src/androidTest/java/com/github/se/di/module/TestModule.kt
```diff
@@ -158,7 +159,7 @@ object MockFirebaseFirestoreModule {
         .thenReturn(mockProfilesQuery)
     `when`(mockProfilesQuery.whereLessThanOrEqualTo(anyString(), anyString()))
         .thenReturn(mockProfilesQuery)
-    `when`(mockProfilesQuery.get()).thenReturn(mockTaskQuerySnapshotProfiles)
+    `when`(mockProfilesQuery.get(Source.SERVER)).thenReturn(mockTaskQuerySnapshotProfiles)
```
MaxCastelCast, line 162: Question : I'm curious, why do you have to add the Source.SERVER here ? What does it represent ?

MaxCastelCast (changes requested):
Hello ! This PR is very good, it is a bit lightweight but since it is the last week, it is better to avoid introducing bugs at last moment. Botond has already done a lot of work for the team and the main features have already been introduced, just fix the issues and i ll be happy to approve your PR

app/src/androidTest/java/com/github/se/di/module/TestModule.kt
BotondAKovacs: Basically that's why the offline mode was broken, it was removed by someone at some point. The idea is that for for our reconnection logic to launch, get() has to return onFailure, but firebase is well made and if there is no internet it gets the information from cache so it never launches the reconnection logic, by adding this you force firebase to get the information from the server which in turn, when fails, launches our reconnection logic with it's onFailure

app/src/main/java/com/github/se/icebreakrr/MainActivity.kt
BotondAKovacs: We can discuss

MaxCastelCast (commented):
I tested your code locally and the notifications work greatly, well done !

app/src/androidTest/java/com/github/se/di/module/TestModule.kt
MaxCastelCast: Ok it is clearer now, thank you !

MaxCastelCast (approved):
Looks good to me 👍

PR #286, opened 2024-12-18
Feature/meeting request bug fixes

MaxCastelCast:
Fixes the following bugs :
- The markers and the textes under the markers are too far away in the LocationViewMap
- A toast is showed when you accept a meeting request (should be remove since it is deprecdated)
- When clicking on a marker in the MapScreen, the message is odd "accepted/cancelled meeting request" , should be replaced
- Remove the timeout cancellation for meeting request since it isn't useful for the app anymore and could cause some unexpected bugs

The new LocationViewMap : 
image

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
@@ -128,18 +128,20 @@ class MeetingRequestService : FirebaseMessagingService() {
         val stringReason =
             when (message) {
               "DISTANCE" -> DISTANCE_REASON_CANCELLATION
-              //"TIME" -> TIME_REASON_CANCELLATION
+              // "TIME" -> TIME_REASON_CANCELLATION
               "CANCELLED" -> CANCELLED_REASON_CANCELLATION
               "CLOSED" -> CLOSED_APP_REASON_CANCELLATION
               else -> DEFAULT_REASON_CANCELLATION
             }
-        if(stringReason != DEFAULT_REASON_CANCELLATION) {
+        if (stringReason != DEFAULT_REASON_CANCELLATION) {
           MeetingRequestManager.meetingRequestViewModel?.removeFromMeetingRequestSent(senderUid) {
-            MeetingRequestManager.meetingRequestViewModel?.removeFromMeetingRequestInbox(senderUid) {
-              MeetingRequestManager.meetingRequestViewModel?.removeChosenLocalisation(senderUid) {
-                MeetingRequestManager.meetingRequestViewModel?.updateInboxOfMessages {}
-              }
-            }
+            MeetingRequestManager.meetingRequestViewModel?.removeFromMeetingRequestInbox(
+                senderUid) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
```diff
@@ -128,21 +128,24 @@ class MeetingRequestService : FirebaseMessagingService() {
...
+            MeetingRequestManager.meetingRequestViewModel?.removeFromMeetingRequestInbox(
+                senderUid) {
+                  MeetingRequestManager.meetingRequestViewModel?.removeChosenLocalisation(
+                      senderUid) {
+                        MeetingRequestManager.meetingRequestViewModel?.updateInboxOfMessages {}
+                      }
+                }
           }
+          MeetingRequestManager.meetingRequestViewModel?.stopMeetingRequestTimer(senderUid, this)
+          showNotification("Cancelled meeting with $senderName", stringReason)
```
BotondAKovacs, line 147: Hardcoded string?

BotondAKovacs (commented):
LGTM, just a question.

Can you elaborate in why we need to remove the timeout?

app/src/main/java/com/github/se/icebreakrr/model/message/MeetingRequestService.kt
MaxCastelCast: Yes, the problem is that we have the variable senderName in the string, I could add a constant for "Cancelled meeting with" and then concatenate the two strings, but i felt that this way it is cleaner, what do you think ?

BotondAKovacs: I'm ambivalent. At the same time I understand you're rational but still it's a String that the users see that fits in strings.xml. Do as you think is best.

MaxCastelCast:
About the first question : The thing is that sometimes, when i tested with you or other people, the meeting request timeout was randomely triggered for some reason even if it worked well on my emulator. I think when adding new fitures (like the cancellation when the user closes the app) is a better and a more elegant solution for cancellations than a timeout. And the 20 minute timeout became obselete since now we have the closed app cancellation, in my opinion

MaxCastelCast:
For your second question: of cours ! 
Before : 
image
After : 
image

BotondAKovacs:
> For your second question: of cours ! Before : image After : image

Thank! Just one last question, have you tested if it looks better with a slight gap between the box and the pin?

BotondAKovacs:
> About the first question : The thing is that sometimes, when i tested with you or other people, the meeting request timeout was randomely triggered for some reason even if it worked well on my emulator. I think when adding new fitures (like the cancellation when the user closes the app) is a better and a more elegant solution for cancellations than a timeout. And the 20 minute timeout became obselete since now we have the closed app cancellation, in my opinion

fair enough

MaxCastelCast:
Yes ! It is indeed a good idea ! 
image

BotondAKovacs (approved):
LGTM, please just put a screenshot of the new UI in the summary.

PR #287, opened 2024-12-18
Bugfix/274

arthur-mrgt:
# Prevent crash when returning to app after removing location permissions

This PR addresses the issue where the app crashes when location permissions are removed, and the user returns to the app.  

### Changes  
- **Added `hasRequiredPermissions`:**  
  - A new function to verify `ACCESS_FINE_LOCATION` and `FOREGROUND_SERVICE_LOCATION` permissions.  
  - Ensures the service only runs with required permissions.  

- **Updated `onStartCommand`:**  
  - Integrated `hasRequiredPermissions` to stop the service gracefully if permissions are missing.  
  - Prevents crashes caused by missing permissions.  

- **Refactored `ILocationService`:**  
  - Removed `createNotificationChannel` from the interface and moved it to internal scope.  

- **Improved Tests:**  
  - Added tests to validate behavior of `hasRequiredPermissions` for various permission states.  

### Remark  
- The `PermissionManager` was not used in this implementation due to the difficulty of injecting it given the way `LocationService` is initialized. Additionally, using `PermissionManager` would have made testing more complex. For a simple use case like this, the `hasRequiredPermissions` function is straightforward, effective, and well-suited.  

### Testing  
- Verified behavior with and without location permissions.  
- Ensured the app does not crash and handles permission changes gracefully.

### DIFF
--- a/app/src/main/java/com/github/se/icebreakrr/model/location/ILocationService.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/location/ILocationService.kt
@@ -9,6 +9,4 @@ interface ILocationService {
   ): Boolean
 
   fun stopLocationUpdates()
-
-  fun createNotificationChannel()
 }
--- a/app/src/main/java/com/github/se/icebreakrr/model/location/LocationService.kt
+++ b/app/src/main/java/com/github/se/icebreakrr/model/location/LocationService.kt
@@ -1,16 +1,21 @@
 package com.github.se.icebreakrr.model.location
 
+import android.Manifest
 import android.annotation.SuppressLint
 import android.app.Notification
 import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.app.Service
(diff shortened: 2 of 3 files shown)


BotondAKovacs (changes requested):
I went on your branch, launched the app, put it in background, went to the sitting, deactivated location permissions, came back and the app keeps crashing

arthur-mrgt:
I’ve tested extensively with all possible scenarios, and the app no longer crashes on my side.

Are you certain the crash was caused by the issue I worked on and not something else?
Did you encounter a specific fatal error related to missing permissions?

The app checks for the required permissions before performing any sensitive actions, so I don’t see how it could still crash at this point.

BotondAKovacs (approved):
It's weird, it doesn't crash anymore, my emulator may have been bugged last night, maybe it was on a previous version. 
LGTM

